PR #126, opened 2023-05-19
draft;light-prover-sdk

Shigoto-dev19:
Changes:
- added a new dir under the title of light-prover-sdk.
- added stand-alone generic types for a circuit proof & public inputs as long as an IDL is available including the parsed zk account structs.
- The generics.ts file also includes many other types that enable fetching circuit names, and accounts, and parsing to dimensional arrays of strings or numbers, all depending on usage.
- Also added a draft of how these types can make the prover class a generic class specific to the circuit name.

Notes:
- I did not run the tests locally.
- Generic Prover class should be adapted in case to be adopted.


light-prover-sdk/tests/genericProver.test.ts
```diff
@@ -0,0 +1,353 @@
...
+    deposit_utxo1,
+    relayer,
+    keypair,
+    paramsDeposit,
+    paramsWithdrawal;
+
+  before(async () => {
+    poseidon = await circomlibjs.buildPoseidonOpt();
+    // TODO: make fee mandatory
+    relayer = new Relayer(
```
ananas-block, line 52: why do you need a relayer in this test it should work without a blockchain then you don't need a relayer

light-prover-sdk/tests/genericProver.test.ts
```diff
@@ -0,0 +1,353 @@
...
+      mockPubkey3,
+      mockPubkey,
+      mockPubkey,
+      new anchor.BN(5000),
+    );
+    keypair = new Account({ poseidon: poseidon, seed: seed32 });
```
ananas-block, line 58: shouldn't be necessary

light-prover-sdk/tests/genericProver.test.ts
```diff
@@ -0,0 +1,353 @@
...
+      inputUtxos: [deposit_utxo1],
+      transactionMerkleTreePubkey: mockPubkey2,
+      poseidon,
+      recipientSpl: mockPubkey,
+      recipientSol: lightProvider.wallet?.publicKey,
+      action: Action.UNSHIELD,
+      relayer,
+      transactionNonce: 0,
+      verifierIdl: IDL_VERIFIER_PROGRAM_ZERO,
+    });
```
ananas-block, line 101: essentially most if not all of the before statement should not be necessary for this test

light-prover-sdk/tests/genericProver.test.ts
```diff
@@ -0,0 +1,353 @@
...
+  });
+
+  it("prover functional test1", async () => {
+    let tx = new Transaction({
```
ananas-block, line 105: ah now I get why you added the before statement.
I think we should not test transaction compatibility here but in zk.js, because zk.js imports this sdk. Actually I am not sure why this even works since you will have a circular dependency between zk.js and the prover-sdk.

- please test the generic prover for a non light protocol circuit
- remove the light protocol specific tests this should be covered in zk.js tests

light-prover-sdk/src/generics.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export type VerifierIdls = 
+    typeof IDL_VERIFIER_PROGRAM_ZERO |
+    typeof IDL_VERIFIER_PROGRAM_ONE |
+    typeof IDL_VERIFIER_PROGRAM_TWO |
+    typeof IDL_VERIFIER_PROGRAM_STORAGE |
+    typeof IDL_MERKLE_TREE_PROGRAM;
+
+// type IDL0 = typeof IDL_VERIFIER_PROGRAM_ZERO;
+// type IDL2 = typeof IDL_VERIFIER_PROGRAM_TWO;
```
ananas-block, line 17: pls remove unnecessary comments

ananas-block (changes requested):

light-prover-sdk/src/generics.ts
Shigoto-dev19: I will keep your feedback in mind, for a future refactoring of prover generics.

PR #134, opened 2023-05-23
cli: bumped version to alpha.10

ananas-block:
- released light cli alpha 10


SwenSchaeferjohann (approved):
Just noting here as well that .10 release doesn't exist rn

PR #143, opened 2023-06-06
zk.js: fix merkle tree update

ananas-block:

PR #144, opened 2023-06-06
zk.js: make provider reusable

ananas-block:
this pr builds on  https://github.com/Lightprotocol/light-protocol/pull/143 and should be merged after

Problem:
- one anchor provider cannot be used to instantiate a merkle tree twice because it will send the same transactions again
- in case of the compute transactions the exact same transaction since the different data is just saved in the temporary account

Solution:
- instantiate a new connection and provider when updating the Merkle tree with the test relayer
- changed the user_test to test the new changes by only instantiating the provider and user once

Comment:
- this is somewhat a quick fix, we should think about a better solution maybe change the Merkle tree program to make transactions unique

### DIFF
--- a/light-system-programs/tests/user_tests.ts
+++ b/light-system-programs/tests/user_tests.ts
@@ -36,7 +36,7 @@ import {
 import { BN } from "@coral-xyz/anchor";
 
 var POSEIDON;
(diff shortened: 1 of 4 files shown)


vadorovsky (approved):
Let's hope that rebase will make it green

PR #155, opened 2023-06-15
zk.js: user - unshield unified sol and spl recipient

ananas-block:
Problem:
- unshield recipient arguments were confusing

Solution:
- take a recipient sol account and use the associated token account as destination for the spl unshield

### DIFF
--- a/light-zk.js/src/test-utils/testStateValidator.ts
+++ b/light-zk.js/src/test-utils/testStateValidator.ts
@@ -13,6 +13,7 @@ import {
   TOKEN_REGISTRY,
 } from "../constants";
 import { Utxo } from "../utxo";
+import { MINT } from "./constants_system_verifier";
 
 export type TestInputs = {
   amountSpl?: number;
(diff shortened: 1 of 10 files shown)


light-zk.js/src/test-utils/testStateValidator.ts
Shigoto-dev19, line 1: - code quality wise, it would be better to assign balance when undefined as the following: 
=> the ?? operator assign an alternative value in case of undefined

          try {
            var balance = (
              await this.provider.provider?.connection.getTokenAccountBalance(
                userBalances.splAccount,
              )
            )?.value.uiAmount?? 0;
          } catch (error) {
          userBalances.preTokenBalance = balance;

SwenSchaeferjohann (approved):

Shigoto-dev19 (approved):
- The changes to airdrop.ts will solve the confusion for the user to assign the authority public key when trying to airdrop spl using mintTo method from the sdk.
- All in all, the code looks more readable and more straightforward.
- One thing I forgot to mention is that an error handling for the airdrop of unregistered SPL tokens should be added in coordination with Shield of course.

ananas-block:
Good point on airdrop, it’s just for testing so error handling is less critical here but we should keep that in mind in general.

PR #156, opened 2023-06-15
merkle tree program - merkle tree update: removed logs

ananas-block:
Problem:
- https://github.com/Lightprotocol/light-protocol/actions/runs/5281318114/jobs/9554721727?pr=145 ci tests failed because it ran out of compute units during merkle tree update
Solution:
- removed logs, which is lowest hanging fruit to save compute units


Shigoto-dev19 (approved):
clean solution!, everyting looks fine

PR #157, opened 2023-06-16
zk.js: merkle tree update added retry logic

ananas-block:
Problem:
- merkle tree update is still flaky

Solution Attempt:
- did not crash during a 6h test in ci so hopefully this is a sustainable solution

- refactored merkle tree update compute instruction execution function
  - send 28 transactions (the exact amount of transactions to update the Merkle tree when inserting 1 pair of leaves)
  - check whether Merkle tree update is complete
  - if not send another 10 transactions
  - the update might not be complete after 28 transactions because it is a batched update the update (for more insertions we need to compute more hashes -> more transactions) or transactions failed
  - if a limit of 120 transactions is reached an error is thrown


SwenSchaeferjohann (approved):
lgtm

PR #158, opened 2023-06-16
zk.js: added confirm options

ananas-block:
Problem:
- user shield/transfer/unshield wait for merkle tree to update which takes a couple seconds while the transaction is already finalized

Solution:
- add two confirm options:
  - finalized (utxos are spend and new utxos are queued but not inserted into the merkle tree yet)
    -> transaction is final but new utxos are not spendable yet
  - spendable (utxos are spent and new utxos are spendable -> have been inserted into the merkle tree)

### DIFF
--- a/light-system-programs/tests/user_tests.ts
+++ b/light-system-programs/tests/user_tests.ts
@@ -1,5 +1,5 @@
 import * as anchor from "@coral-xyz/anchor";
-import { Keypair as SolanaKeypair } from "@solana/web3.js";
+import { Keypair as SolanaKeypair, PublicKey } from "@solana/web3.js";
 import _ from "lodash";
 const chai = require("chai");
 const chaiAsPromised = require("chai-as-promised");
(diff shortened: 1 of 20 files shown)


light-zk.js/src/wallet/user.ts
```diff
@@ -60,7 +61,10 @@ import { Idl } from "@coral-xyz/anchor";
 const message = new TextEncoder().encode(SIGN_MESSAGE);
 
 // TODO: Utxos should be assigned to a merkle tree
-// TODO: add support for wallet adapter (no access to payer keypair)
+export enum ConfirmOptions {
```
SwenSchaeferjohann, line 64: Should we move this to a separate file? /config or constants file

SwenSchaeferjohann (approved):
Given that ci failure is unrelated ; minor comments questions , but non blocker so feel free to merge when looked thru the comments

light-zk.js/src/wallet/user.ts
ananas-block: in general yes, but I also don't like a lot of files with 4 lines

light-system-programs/tests/user_tests.ts
```diff
@@ -189,6 +192,99 @@ describe("Test User", () => {
...
+
+    const recipientSeed = bs58.encode(new Uint8Array(32).fill(8));
+    const recipientUser: User = await User.init({
+      provider,
+      seed: recipientSeed,
+    });
+
+    let testInputsTransfer = {
+      amountSpl: 1,
+      token: "USDC",
```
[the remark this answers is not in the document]
ananas-block: how do you handle it in the ui for tokens?

PR #163, opened 2023-06-25
Add testnet support

ananas-block:
Changes:
- updated program ids to testnet
- updated accounts pubkeys and json files for tests which depend on program ids
- replaces .rpc calls in MerkleTreeConfig calls with sendAndConfirmTransaction - had to adjust the merkle tree tests for different error propagation
- added ts script for 100 sol airdrops from testnet
- added ts script which instantiates a look up table - can be deleted once cli is ready


light-zk.js/src/relayer.ts
```diff
@@ -107,7 +107,7 @@ export class Relayer {
     provider: Provider,
   ): Promise<RelayerSendTransactionsResponse> {
     try {
-      const response = await axios.post(this.url + "/relayInstruction", {
+      const response = await axios.post(this.url + "/relayTransaction", {
```
SwenSchaeferjohann, line 110: arent we technically still sending instruction**s** ?

spl_noop.so
SwenSchaeferjohann, line 1: why push .so files ?

SwenSchaeferjohann (commented):
besides the comments lgtm! feel free to merge after addressing the comments

also generally wondering how % test coverage is here?

light-system-programs/scripts/airdropTestnet.ts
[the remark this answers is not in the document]
ananas-block: this is how I funded the testnet deployment.
It’s not fast. The loop sleeps for 10s after every airdrop to avoid ratelimiting.

light-zk.js/src/relayer.ts
ananas-block: Yes we are sending several instructions which in concert execute one shielded transaction.
Every instruction is executed in a single solana transaction.

happy for suggestions on better naming!

spl_noop.so
ananas-block: I was struggling with failing tests for mock verifier.
Apparently the docker image didn’t use the local verifier program two binary.
Therefore, I changed the runTest.sh script to start a test validator without docker. But the spl noop binary has been saved in the docker image so far.
It was easier to push this than to change the docker image.

ananas-block:
> besides the comments lgtm! feel free to merge after addressing the comments
> 
> also generally wondering how % test coverage is here?

test coverage of what specifically?

(Most of the changes are to code which is already covered by tests and for the scripts to init a lookup table and airdrop tests are overkill in my opinion because it is a temporary solution these should be part of the cli.)

light-zk.js/src/relayer.ts
SwenSchaeferjohann: probably `relayInstructions` makes most sense then? but this is not urgent so can still merge this pr now

spl_noop.so
SwenSchaeferjohann: okay I see, so this is probably temporary?

PR #165, opened 2023-06-26
zk.js: added payer to send versioned transaction

ananas-block:
- removed double spl_noop.so and adapted runTest.sh scripts
- send versioned transaction:
  - solved issue that the relayer pubkey was added as signer even if it is different from the signer during shield
  - added payer to be more explicit about the signer and payer of the transaction
  - the payer variable also unifies the function for both the test relayer and transaction class which uses the function directly to shield

Note:
- our tests are really ugly, will refactor in a different pr
  - looking to unify (the merge tests are distributed to stay under the 10 transaction threshold)
  - will unify test setup
  - sol unshield and transfer tests are missing


SwenSchaeferjohann (approved):
lgtm

PR #166, opened 2023-06-26
zk.js - utxo encryption

ananas-block:
Problem:
- transaction nonce as domain separator is unreliable because it is easy to produce situations in which the transaction nonce is not up to date

Solution:
- changed domain separator from transaction nonce to commitment hash
- removed transaction nonces completely

Note:
- this will be revisited when benchmarking performance of encryption
- we should also evaluate whether we want to switch from blake2b to a hdkf


SwenSchaeferjohann (approved):
Lgtm

PR #167, opened 2023-07-03
user: add randomized tests & refactor indexer

ananas-block:
### Randomized Tests:
- added randomized test in light system programs:
- three modes sol only, spl only, both
- these tests are not intended to run in ci but are only for manual testing
Test setup:
- create a relayer from fresh keypairs
- initialize a number of users
- in an infinite loop do:
  - select random user
  - select random token if both
  - select random action
  - merge utxos in case there are utxos in the inbox
  - execute action
 
### indexRecentTransaction:
- streamlined processIndexedTransaction
  - find the inner transaction which calls one of the system verifiers
  - in the instruction the order of account keys is deterministic
  - thus once we have the instruction we can assign the account keys
  - derive the type and amounts
  - compute change amount (to be consistent with solana explorer)
- adapted the order of account keys in verifiers one, storage, and two to be consistent with verifier zero
- I moved variable accounts which are not part of verifier zero to the bottom of the account struct so that the order of the previous accounts is consistent
  @vadorovsky we need to add the change into the account macro as well.

### DIFF
--- a/light-system-programs/tests/user-random.tests.ts
+++ b/light-system-programs/tests/user-random.tests.ts
@@ -21,6 +21,13 @@ import {
 } from "@lightprotocol/zk.js";
 
 import { BN } from "@coral-xyz/anchor";
+import {
+  transfer as splTransfer,
+  getAccount,
+  getAssociatedTokenAddressSync,
+  createAssociatedTokenAccount,
+} from "@solana/spl-token";
+import { assert } from "chai";
 
 function generateRandomTestAmount(
   min: number = 0.2,
@@ -105,7 +112,7 @@ const shield = async (
     testInputs.token,
   );
 
-  const testStateValidator = new UserTestAssertHelper({
+  const userTestAssertHelper = new UserTestAssertHelper({
     userSender: user,
     userRecipient: user,
     provider: user.provider,
@@ -125,7 +132,7 @@ const shield = async (
       keypair,
     );
   }
-  await testStateValidator.fetchAndSaveState();
+  await userTestAssertHelper.fetchAndSaveState();
 
   const res = await user.shield({
     publicAmountSol: testInputs.amountSol,
@@ -137,9 +144,9 @@ const shield = async (
   await user.provider.latestMerkleTree();
 
   if (isSol) {
-    await testStateValidator.checkSolShielded();
+    await userTestAssertHelper.checkSolShielded();
   } else {
-    await testStateValidator.checkSplShielded();
+    await userTestAssertHelper.checkSplShielded();
   }
 };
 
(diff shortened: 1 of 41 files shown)


SwenSchaeferjohann:
Hypothetically, if we want to index custom PSPs does the system work even if devs change the order of account keys?

ananas-block:
> Hypothetically, if we want to index custom PSPs does the system work even if devs change the order of account keys?

we only look at the instructions which call a system verifier, psp devs cannot change the order of the accounts for those instructions.

light-system-programs/tests/user-random.tests.ts
```diff
@@ -0,0 +1,473 @@
...
+  });
+
+  /**
+   * fails at tx 22
+   * - had two shields before one usdc and one sol
+   * - history only finds the lastest sol shield
+   *
+   * Check:
+   * - that usdc transactions are categorized correctly
+   * */
```
SwenSchaeferjohann, line 469: I suppose this is wip?

light-zk.js/src/constants.ts
```diff
@@ -61,6 +61,13 @@ export const LOOK_UP_TABLE = new PublicKey(
   "DyZnme4h32E66deCvsAV6pVceVw8s6ucRhNcwoofVCem",
 );
 
+export const MAX_U64 = new anchor.BN("18446744073709551615");
+export const VERIFIER_PUBLIK_KEYS = [
```
SwenSchaeferjohann, line 65: Typo publik

light-zk.js/src/transaction/indexRecentTransactions.ts
```diff
@@ -136,88 +163,101 @@ async function processIndexedTransaction(
...
+    publicAmountSpl: Uint8Array,
+    publicAmountSol: Uint8Array,
+  ) => {
+    let type = Action.SHIELD;
+    let amountSpl = new BN(publicAmountSpl, 32, "be");
+    let amountSol = new BN(publicAmountSol, 32, "be");
+
+    let splIsu64 = amountSpl.lte(MAX_U64);
+    let solIsu64 = amountSol.lte(MAX_U64);
+    if (!splIsu64 || !solIsu64) {
```
SwenSchaeferjohann, line 198: Wouldnt it be solIsU64 with capitalised U?

SwenSchaeferjohann (commented):
Lgtm would like @vadorovsky taking a look too given the scope of the PR

light-zk.js/src/constants.ts
ananas-block: lol

light-system-programs/tests/user-random.tests.ts
ananas-block: yes, and the tx history is out of scope for this pr

SwenSchaeferjohann (approved):
lgtm

PR #169, opened 2023-07-04
prover.ts: decouple prover.ts from zk.js

sergeytimoshin:
Draft based on #126 

* Removed all redudant tests from light-provers-sdk/tests, since they already not really generic and already implemented in zk.js (https://github.com/Lightprotocol/light-protocol/blob/main/light-zk.js/tests/prover.test.ts)
* A few cleaned up things here and here.


ananas-block:
Thanks!

I noticed that it is not used in zk.js yet (it probably wasn't used in the prior pr either).
In zk.js it is used in [transaction.ts](https://github.com/Lightprotocol/light-protocol/blob/main/light-zk.js/src/transaction/transaction.ts#L377)
Can you replace the prover class in zk.js?
(Then it will be naturally included in more tests in zk.js and light-system programs as well.)

light-prover-sdk/tests/genericProver.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    assert.equal(
+      lightProvider.solMerkleTree?.merkleTree.indexOf(
+        deposit_utxo1.getCommitment(poseidon),
+      ),
+      0,
+    );
+  });
+
+  it("Test Generic Self-contained Prover for VerifierZero", async () => {
```
ananas-block, line 72: regarding tests you could add a couple more standard tests:
- failing
- run full prove two times and receive the same public inputs as output (the proof will be different ofc because it is using different randomness)

light-prover-sdk/src/generics.ts
```diff
@@ -0,0 +1,130 @@
...
+// Extract unique circuit names from VerifierIdls Union Type
+export type CircuitNames = ExtractPrefix<ZKAccounts['name']>;
+
+// Optional type: circuit proof inputs object by selecting full zk account name from the idl
+type ZKProofInputsObjectFullName<Idl extends VerifierIdls, AccountName extends ZKAccounts['name']>  =
+    CircuitInputsObject<MapObjectKeys<SelectZKAccount<Idl, AccountName>>>;
+
+
+// let sampleProofInputs: ProofInputs<IDL2, "transactionApp4">;
+// sampleProofInputs.
```
ananas-block, line 68: left over comments I suppose

light-prover-sdk/src/generics.ts
```diff
@@ -0,0 +1,130 @@
...
+export type ProofInputs<Idl extends VerifierIdls, CircuitName extends CircuitNames>  =
+    CircuitInputsObject<MapObjectKeys<SelectZKAccount<Idl, `zK${CircuitName}ProofInputs`
+    >>>;
+
+export type ParsedPublicInputs<Idl extends VerifierIdls, CircuitName extends CircuitNames>  =
+    CircuitParsedPubInObject<MapObjectKeys<SelectZKAccount<Idl, `zK${CircuitName}PublicInputs`
+    >>>;
+
+// let publicInputs: ParsedPublicInputs<IDL2, "transactionApp4">;
+// publicInputs.
```
ananas-block, line 79: left over comments I suppose

ananas-block:
snarkjs's process does not terminate automatically can you add the following to terminate it?
```
(globalThis as any).curve_bn128.terminate();
```

see https://github.com/iden3/snarkjs/issues/152

light-prover.js/tests/prover.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    assert.equal(
+      lightProvider.solMerkleTree?.merkleTree.indexOf(
+        deposit_utxo.getCommitment(poseidon),
+      ),
+      0,
+    );
+  });
+
+  after(async () => {
+    globalThis.curve_bn128.terminate();
```
ananas-block, line 76: My intuition was to put this into the prove method.
Do you see tradeoff putting it in there?

PR #172, opened 2023-07-07
zk.js - user - unshield: fixed associated token account creation

ananas-block:
Problem:
- unshielding to a sol address which already has an initialized associated token account fails because it always tries to create a new token account

Solution:
- fix detection whether an associated token account already exists, thus don't try to create one when it already exists

### DIFF
--- a/light-zk.js/src/wallet/user.ts
+++ b/light-zk.js/src/wallet/user.ts
@@ -724,14 +724,9 @@ export class User {
         tokenCtx!.mint,
         recipient,
       );
-
-      let tokenBalance =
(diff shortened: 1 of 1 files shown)


SwenSchaeferjohann (commented):
are we sure getTokenAccountBalance doesn't throw if not account was instantiated before?

ananas-block:
> are we sure getTokenAccountBalance doesn't throw if not account was instantiated before?

thanks for the remark. I replaced it with getAccountInfo.
tests caught it as well https://github.com/Lightprotocol/light-protocol/actions/runs/5490795921/jobs/10006664446

SwenSchaeferjohann (approved):

PR #173, opened 2023-07-08
Swen indexer db

SwenSchaeferjohann:
adds
- relay endpoint (/relayTransaction) that can be hosted locally or remote 
- index endpoint (/indexedTransactions)
- relay job queue and concurrent workers 
- uses redis (bullmq job queue library) for both relay and indexing (at some later point we'd want to move the indexing parts to a separate non-job queue setup for better scalability)
- renamed parsing functions for better clarity around the e2e indexing process in (/indexerService/fetchRecentTransactions.ts)

**doesn't touch**
- any logic wrt the parsing itself

### Jorrit Additions:

- in runtTest.sh - added killall to start regardless whether an existing test-validator is already running
- zk.js: removed lookupTable from Relayer object
- refactored parts of the provider to make lookup table management easier
- light-system-programs - tests: added separate relayer account
- added redis to the install script depending on an argument
- storing amounts as hexstrings in IndexedTransaction objects, which are parsed in the client to ParsedIndexedTransaction objects


SwenSchaeferjohann:
- zkjs: indexedTransaction type now converts BN's (e.g. pubAmounts) to hexstrings and respectively does a new BN().toNumber instead of just toNumber() wherever they are used. this is to ensure consistency between testRelayer and relayer implementations (over wire the bn is always stringified)

light-zk.js/src/test-utils/initLookUpTable.ts
```diff
@@ -137,21 +139,21 @@ export async function initLookUpTableTest(
...
       }
     }
 
-    const extendInstruction = AddressLookupTableProgram.extendLookupTable({
-      lookupTable: lookupTableAddress,
-      authority: payerPubkey,
-      payer: payerPubkey,
-      addresses: addressesToAdd,
-    });
+    // const extendInstruction = AddressLookupTableProgram.extendLookupTable({
```
vadorovsky, line 149: Any reason it's commented, but not removed?

relayer/.env.example
```diff
@@ -1,4 +1,13 @@
...
 RELAYER_RECIPIENT=[ 79, 149, 114, 205, 153,  39, 179,  25,  52, 107, 164, 75, 124, 102,  73,  91, 112,  61, 152,  70, 250, 208, 134,  24, 159, 176,  52, 177, 183,  99, 151, 146, 140, 232,  23, 147,  47,  68, 206,  56, 135, 249,  23,  34, 252, 245, 228,  51, 240,  38, 154, 200, 137,   2, 166, 173,  64, 131, 141,  14, 150, 134,  80,  13 ]
-LOOK_UP_TABLE=""
-TEST_ENVIROMENT=true
\ No newline at end of file
+LOOK_UP_TABLE="4Dcx88YhY6YD4ojbGtK1e1x344WPhurHM6GxX1JCDkAU"
+TEST_ENVIRONMENT=true
+ENVIRONMENT=LOCAL
+NETWORK=LOCAL
+RPC_URL="http://localhost:8899"
+# DATABASE_URL=${mainnet-testing-v1.DATABASE_URL}
```
vadorovsky, line 8: Why is it commented?

relayer/src/services/indexerService/indexer.ts
```diff
@@ -0,0 +1,95 @@
...
+  job,
+  token,
+}: {
+  connection: Connection;
+  newRawTransactions: TransactionResponse[];
+  job: Job;
+  token: any;
+}) {
+  const newParsedTransactions: IndexedTransaction[] = [];
+  //   await Promise.all(
```
vadorovsky, line 19: Can we remove it?

scripts/install.sh
```diff
@@ -204,4 +204,17 @@ download_file_github \
     macro-circom \
     "${PREFIX}/bin"
 
+if $1 == "true" ; then
```
vadorovsky, line 207: I don't think passing a boolean argument to the script is the cleanest solution - when you see `./install.sh true` written in some other script, it doesn't really give you an idea what's the purpose of `true`.

I would rather have a CLI argument like `--enable-redis` or so. Here is an example how you can parse arguments like that in shell: https://github.com/Lightprotocol/install/blob/main/light-protocol-install.sh#L56-L83

vadorovsky (changes requested):

light-zk.js/src/test-utils/initLookUpTable.ts
ananas-block: Yeah I discovered we have two functions which create lookup tables. I want remove this one in another pr.

relayer/src/services/indexerService/indexer.ts
ananas-block: This is a hook which is not used yet.
@SwenSchaeferjohann what do you think?

scripts/install.sh
ananas-block: Good idea, will add the flag —enable-redis

relayer/.env.example
ananas-block: I suppose this is the url we will use with a separate digital ocean database instance which is not tested yet.
Cc @SwenSchaeferjohann

.github/actions/setup-and-build-with-relayer/action.yml
```diff
@@ -0,0 +1,24 @@
+name: Setup and build with relayer
+description: Checkout sources, install dependencies, build and prepare for tests
+runs:
+  using: "composite"
+  steps:
+    - name: Install dependencies
+      shell: bash
+      run: |
+        ./scripts/install.sh true
```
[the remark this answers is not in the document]
ananas-block: True I will ask :D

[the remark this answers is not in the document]
ananas-block: done

scripts/install.sh
ananas-block: done

relayer/.env.example
SwenSchaeferjohann: yup

SwenSchaeferjohann: yes

relayer/src/services/indexerService/indexer.ts
SwenSchaeferjohann: yeah let's remove it. no big deal to add it later on

relayer/.env.example
SwenSchaeferjohann: I will add more detailed inline comments for this file in a separate PR.

mock-app-verifier/tests/appUtxo.test.ts
SwenSchaeferjohann, line 1: why was this test file deleted?

relayer/src/services/indexerService/index.ts
```diff
@@ -0,0 +1,4 @@
+// export * from "./search"
```
SwenSchaeferjohann, line 1: this can be removed

mock-app-verifier/tests/appUtxo.test.ts
ananas-block: Tests are duplicate with functional tests. Also it was not maintained and executed during ci tests

relayer/src/services/indexerService/indexer.ts
ananas-block: removed

vadorovsky (approved):

PR #178, opened 2023-07-12
Matteo utxo refactoring for speed

xonoxitron:


light-zk.js/src/utxo.ts
```diff
@@ -392,6 +395,7 @@ export class Utxo {
         ).fill(0),
       ]);
       includeAppData = false;
+      // @matteo: why not performing the check below before the array initialization above? we can save some time/space
```
ananas-block, line 398: The usage pattern to create random empty utxos, new Utxo(poseidon) would not work anymore.
However, there is no reason to apply this patter to program utxos (every utxo with app data is a program utxo).
Therefore, I added this check as a safeguard to prevent the creation of program utxos with a random account.

Actually it would be safer if we create a check like this for the case that an amount in amounts is not zero.

xonoxitron: gotcha.
honestly speaking I have to mature a bit better understanding to state if makes sense or not.
I'll ruminate on it!
PS: from your perspective, where this check should happen?

ananas-block: can be on top of the constructor (gut feel)

xonoxitron: cool, thanks. will look into 😎🤙

xonoxitron:
this #178 is the root branch where I started refactoring.
#179 and #180 will be merged as follow-up

ananas-block (approved):

PR #180, opened 2023-07-13
Matteo utxo extra conditions

xonoxitron:
added extra conditions for user balance safety


light-zk.js/src/utxo.ts
```diff
@@ -219,14 +219,34 @@ export class Utxo {
...
       this.assetsCircuit = [
         hashAndTruncateToCircuit(SystemProgram.programId.toBytes()),
         hashAndTruncateToCircuit(this.assets[1].toBytes()),
       ];
-    } else if (this.amounts[0].toString() === "0") {
+    } else if (this.amounts[0].isZero()) {
       this.assetsCircuit = [BN_0, BN_0];
+    }
+    // @matteo: adding condition for amounts =! 0
+    else if (!this.amounts[0].isZero()) {
```
ananas-block, line 244: Why do you always throw an error if the amounts[0] is non zero?

amounts[0] is the sol amount of the utxo it's intended to be non zero when used and is allowed to be zero when it is used as a filling utxo.

xonoxitron: made no sense, I commented it

ananas-block: top, tests look good can you delete it?
Or is there a reason why we would want to keep the comment?

xonoxitron: removed!

PR #199, opened 2023-07-26
feat: Started implementing UTXO with prefix hash

xonoxitron:
Finalizing solution implementation

Implemented additional tests for light-zk.js/account.ts

Fixed missing inferred types in Utxo functional test

Fixes

Added some debug logs

Fixed 1 out of 2 tests on UTXO

Fixed 2 out of 2 tests on UTXO

Added failure test to utxo.test.ts


light-zk.js/src/utxo.ts
```diff
@@ -630,27 +635,54 @@ export class Utxo {
...
+    commitment,
+    prefix,
+  }: {
+    account: Account;
+    commitment: Uint8Array;
+    prefix: Uint8Array;
+  }): boolean {
+    return (
+      account.generateUtxoPrefixHash(commitment, PREFIX_LENGTH).join(",") ===
+      prefix.join(",")
```
ananas-block, line 672: interesting :D
did you consciously decide to use join over toString()?

light-zk.js/src/utxo.ts
```diff
@@ -630,27 +635,54 @@ export class Utxo {
...
         .create()
         .update(Uint8Array.from([...nonce, ...bytes_message]))
         .digest();
 
       // adding the 8 bytes as padding at the end to make the ciphertext the same length as nacl box ciphertexts of 120 bytes
-      return Uint8Array.from([...ciphertext, ...padding.subarray(0, 8)]);
+      return Uint8Array.from([
+        ...prefix,
+        ...ciphertext,
+        ...padding.subarray(0, 8),
```
ananas-block, line 651: ah interesting I forgot about the padding. We could also think about replacing this with the prefix, in case we decide we don't need the prefix for asym encryption.

light-zk.js/src/account.ts
```diff
@@ -278,6 +288,26 @@ export class Account {
...
+      .digest();
+  }
+
+  generateUtxoPrefixHash(commitmentHash: Uint8Array, length?: number) {
+    const input = Uint8Array.from([
+      ...this.getDomainSeparatedHashingSecretKey("hashing"),
+      ...commitmentHash,
+    ]);
+
+    return blake2b.create(b2params).update(input).digest().slice(0, length);
```
ananas-block, line 308: instead of slice you could adjust the length in the b2params

light-zk.js/src/account.ts
```diff
@@ -35,7 +34,8 @@ export class Account {
   };
   eddsa: any;
   aesSecret?: Uint8Array;
-
+  hashingSecret?: Uint8Array;
+  prefixHash?: Uint8Array;
```
ananas-block, line 38: what's your reasoning for saving a prefix hash here?

light-zk.js/src/account.ts
```diff
@@ -85,6 +87,7 @@ export class Account {
         secretKey: new Uint8Array(32),
       };
```
ananas-block, line 88: we should add the hashingSecret here as well, I think this is also the only place where the passed in aes secret is used.
Correct me if I am wrong :)

light-zk.js/src/account.ts
```diff
@@ -111,7 +114,13 @@ export class Account {
       this.poseidonEddsaKeypair = Account.getEddsaPrivateKey(
         this.burnerSeed.toString(),
       );
-      this.aesSecret = Account.generateAesSecret(this.burnerSeed.toString());
+      this.aesSecret = Account.generateSecret(
+        this.burnerSeed.toString(),
+        "aes",
+      );
+      this.hashingSecret =
+        hashingSecret ||
```
ananas-block, line 122: what's the reasoning for passing assigning the passed in hashing secret here?
(feels inconsistent with the lines above)

light-zk.js/src/account.ts
```diff
@@ -397,8 +427,8 @@ export class Account {
     return privateKey;
   }
 
-  static generateAesSecret(seed: String, _domain?: string): Uint8Array {
-    const privkeySeed = seed + "aes";
+  static generateSecret(seed: String, domain: String): Uint8Array {
```
ananas-block, line 430: nice!

light-zk.js/src/utxo.ts
xonoxitron: yeah I did. *.join is faster:

Screenshot 2023-07-27 at 09 51 40

light-zk.js/src/utxo.ts
xonoxitron: yeah, I thought about this I remember we talked about this, let's agree on the best option!

light-zk.js/src/account.ts
xonoxitron: cool! I didn't know :)

light-zk.js/src/account.ts
xonoxitron: I mimic-ed the `aesSecret` scope/meaning/behaviour, you've better context if you are convinced that's makes no sense I'll remove it

light-zk.js/src/account.ts
xonoxitron: I'll ruminate on it but I the gut feeing is yes, it makes sense to add it

light-zk.js/src/account.ts
xonoxitron: as per the above, I mimic-ed the `aesSecret` scope/meaning/behaviour

light-zk.js/src/account.ts
xonoxitron: yeah, sprinkles of optimisation here and there :)

light-zk.js/src/account.ts
xonoxitron: after code inspection I agree with you, removed. made no sense to have it

light-zk.js/src/account.ts
xonoxitron: after code inspection -> removed. ✅

light-zk.js/src/utxo.ts
ananas-block: nice :D

light-zk.js/src/account.ts
ananas-block: agreed it's shouldn't be part of the encryption keypair, I just commented at that line because github doesn't allow me to comment above, it's too far from changes

light-zk.js/src/account.ts
xonoxitron: replaced local `b2params` with `{dkLen:PREFIX_LENGTH}` 🚀

light-zk.js/src/account.ts
xonoxitron: cool, makes sense!

light-zk.js/src/account.ts
```diff
@@ -278,6 +287,26 @@ export class Account {
       .digest();
   }
 
+  getUtxoPrefixViewingKey(salt: string): Uint8Array {
+    return this.getDomainSeparatedHashingSecretKey(salt);
+  }
+
+  getDomainSeparatedHashingSecretKey(domain: string): Uint8Array {
+    return blake2b
+      .create({ dkLen: PREFIX_LENGTH })
```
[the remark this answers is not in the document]
xonoxitron: that's a good point. so, what do we encrypt/decrypt with the hashing secret key?
looking at the `getAesUtxoViewingKey` usage it is used in the utxo encrypt/decrypt methods

light-zk.js/tests/utxo.test.ts
```diff
@@ -352,10 +384,49 @@ describe("Utxo Functional", () => {
...
     } else {
-      throw new Error("decrypt failed");
+      throw new Error("decrypt unchecked failed");
+    }
+
+    const receivingUtxoNoAes = new Utxo({
+      poseidon,
+      assets: inputs.assets,
+      amounts: inputs.amounts,
+      account: new Account({ poseidon }),
```
ananas-block, line 397: This is a random account with an aes secret key, thus in line 406 you are encrypting the utxo with aes.

light-zk.js/tests/utxo.test.ts
```diff
@@ -352,10 +384,49 @@ describe("Utxo Functional", () => {
...
+      aes: false,
+      commitment: new anchor.BN(
+        receivingUtxoNoAes.getCommitment(poseidon),
+      ).toBuffer("le", 32),
+      assetLookupTable: lightProvider.lookUpTables.assetLookupTable,
+      verifierProgramLookupTable:
+        lightProvider.lookUpTables.verifierProgramLookupTable,
+    });
+    if (typeof receivingUtxo1NoAes == "boolean") {
+      assert.equal(receivingUtxo1NoAes, false);
```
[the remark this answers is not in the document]
xonoxitron: this is because of this -> https://github.com/Lightprotocol/light-protocol/blob/92c986106169a6dcccf460d3d3fcef0962d9292f/light-zk.js/src/utxo.ts?plain=1#L842-L845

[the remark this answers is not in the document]
xonoxitron: and this -> 
https://github.com/Lightprotocol/light-protocol/blob/92c986106169a6dcccf460d3d3fcef0962d9292f/light-zk.js/src/utxo.ts?plain=1#L826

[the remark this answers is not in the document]
xonoxitron: so in that case we'll have:
a) no UTXO found
b) no collision

so it returns FALSE

light-zk.js/tests/utxo.test.ts
xonoxitron: so do you suggest to use this instead -> 

```
account: Account.fromPubkey(pubKey, poseidon),
```

?

light-zk.js/src/wallet/user.ts
```diff
@@ -186,13 +187,11 @@ export class User {
         const leafLeft = trx.leaves[index];
         const leafRight = trx.leaves[index + 1];
 
+        const w = NACL_ENCRYPTED_COMPRESSED_UTXO_BYTES_LENGTH + PREFIX_LENGTH;
```
ananas-block, line 190: why name the variable ```w```?

light-zk.js/src/wallet/user.ts
```diff
@@ -209,12 +208,7 @@ export class User {
         });
         await decryptAddUtxoToBalance({
           encBytes: Buffer.from(
-            trx.encryptedUtxos.slice(
-              (index / 2) * 240 + 120,
-              (index / 2) * 240 +
-                NACL_ENCRYPTED_COMPRESSED_UTXO_BYTES_LENGTH +
-                120,
-            ),
+            trx.encryptedUtxos.slice(index * w + w, index * w + w + w),
```
ananas-block, line 211:
```suggestion
            trx.encryptedUtxos.slice(index * w + w, index * w + 2 * w),
```

light-zk.js/tests/utxo.test.ts
ananas-block: @sergeytimoshin this is still not solved, like matteo wrote this should be Account.fromPubkey(...)

ananas-block (commented):
Took a first pass will, take another pass tomorrow.

light-zk.js/src/wallet/user.ts
sergeytimoshin: width

light-zk.js/src/utxo.ts
```diff
@@ -627,11 +630,7 @@ export class Utxo {
     compressed: boolean = true,
   ): Promise<Uint8Array> {
     const bytes_message = await this.toBytes(compressed);
-    const commitment = new BN(this.getCommitment(poseidon)).toArrayLike(
-      Buffer,
-      "le",
-      32,
-    );
+    const commitment = new BN(this.getCommitment(poseidon)).toBuffer("le", 32);
```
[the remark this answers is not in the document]
sergeytimoshin: @SwenSchaeferjohann do you remember where was this issue? I think I want to add a new test for this case.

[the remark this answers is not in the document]
SwenSchaeferjohann: can confirm toArrayLike is compatible with Browser while toBuffer is not @sergeytimoshin.

as per the inline docs for BN.toBuffer: 

convert to Node.js Buffer (if available). For compatibility with browserify and similar tools, use this instead: a.toArrayLike(Buffer, endian, length)

light-zk.js/src/utxo.ts
```diff
@@ -791,6 +821,61 @@ export class Utxo {
...
+        commitment,
+        appDataIdl,
+        compressed,
+        assetLookupTable,
+        verifierProgramLookupTable,
+      });
+      if (!utxo)
+        return true; // prefixHash matches but decryption fails so utxo is null -> Returns TRUE (COLLISION)
+      else return utxo; // prefixHash matches and decryption succeeds so utxo is good -> Returns UTXO (VALID)
+    } else return false; // prefixHash doesn't match -> Returns FALSE (NO COLLISION)
```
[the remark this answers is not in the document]
sergeytimoshin: I don't like it either. 
We can revert it back to null or we can convert resulting expression `(Utxo | boolean)` to something more strongly typed like Result<Utxo, UtxoDecryptionError>. 
Matteo used `bool` result  in his collision test, but I'm not sure why this information is important for us, @ananas-block can you comment on this please? https://github.com/Lightprotocol/light-protocol/blob/matteo/utxo-decryption-improvements/light-zk.js/tests/utxo.test.ts#L728
If it's really needed, I'll go with the Result<Utxo, UtxoDecryptionError> refactoring, otherwise I'll just return `null`.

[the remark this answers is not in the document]
ananas-block: - hm, I don't remember the specific reason why we differentiated between the errors and did not just return null.
- In general, I prefer explicit error handling, the more explicit the better.
- In my opinion returning enum instead of a boolean is a good improvement.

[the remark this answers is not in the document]
ananas-block: The collision counter is part of the benchmarking.
Every prefix collision will slow down the process because it means that we tried to decrypt a utxo and werent successful.
https://github.com/Lightprotocol/light-protocol/blob/matteo/utxo-decryption-improvements/light-zk.js/tests/utxo.test.ts#L728

ananas-block (changes requested):
Found a couple small things.

light-zk.js/src/utxo.ts
```diff
@@ -791,6 +820,61 @@ export class Utxo {
     }
   }
 
+  // TODO: unify compressed and includeAppData into a parsingConfig or just keep one
+  /**
+   * @description Decrypts an utxo from an array of bytes by checking the UTXO prefix hash, the last 24 bytes are the nonce.
```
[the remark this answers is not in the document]
ananas-block: @sergeytimoshin the comment hasn't been updated yet

light-zk.js/src/utxo.ts
```diff
@@ -667,33 +666,62 @@ export class Utxo {
...
+      .generateUtxoPrefixHash(commitment, UTXO_PREFIX_LENGTH)
+      .join(",");
+    let p2 = prefixBytes.join(",");
+    return p1 === p2;
+  }
+
   // TODO: unify compressed and includeAppData into a parsingConfig or just keep one
   /**
-   * @description Decrypts a utxo from an array of bytes, the last 24 bytes are the nonce.
+   * @description Decrypts an utxo from an array of bytes without checking the UTXO prefix hash, the last 24 bytes are the nonce.
```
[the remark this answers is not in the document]
ananas-block: @sergeytimoshin the comment hasn't been updated yet

zk.js/src/utxo.ts
```diff
@@ -772,25 +816,100 @@ export class Utxo {
...
+        assetLookupTable,
+        verifierProgramLookupTable,
+      });
+
+      // prefixHash matches but decryption fails so utxo is null -> Returns TRUE (COLLISION)
+      if (utxoResult.value) {
+        return utxoResult;
+      }
+
+      // prefixHash matches and decryption succeeds so utxo is good -> Returns UTXO (VALID)
```
ananas-block, line 900: I will like this comment does not match the code any more, or did I misunderstand?

ananas-block (approved):

PR #202, opened 2023-07-27
fix: align console.log outputs: add spaces before emojis

sergeytimoshin:


cli/src/utils/buildPSP.ts
```diff
@@ -193,7 +193,7 @@ export async function buildPSP(
   console.log("✅ Circom files generated successfully");
 
   const circuitMainFileName = extractFilename(stdout.toString().trim());
-  console.log("🛠️️ Building circuit ", circuitMainFileName);
+  console.log("🛠️️  Building circuit", circuitMainFileName);
```
ananas-block, line 196: don't you think it makes sense to have a space on each side?
(otherwise it might look like this 🛠️️  Building circuitcircuitNameMain)

SwenSchaeferjohann: it automatically adds an empty space between string and variable so it should be fine

SwenSchaeferjohann (approved):

PR #204, opened 2023-07-28
feat: Swen browser env compatibility

SwenSchaeferjohann:
work in progress - do not merge yet

- bash script to copy build-circuit files into browser repos (this is a temporary solution, and we can find a solution for this that doesn't require devs to execute a script manually)
- init anchor provider outside node env 

open
- rebase
- debug lookuptableinit/shield/transfer/unshield fns 
- workerthreads for decryption in browser envs


SwenSchaeferjohann:
please do not merge yet, i just wanna let CI run once

PR #210, opened 2023-08-01
fix: log error stack trace in catch blocks

SwenSchaeferjohann:
- better stack tracing in zkjs


light-zk.js/src/transaction/sendVersionedTransaction.ts
```diff
@@ -68,12 +68,14 @@ export const sendVersionedTransaction = async (
...
+      console.log(error);
+
       retries--;
-      if (retries == 0 || e.logs !== undefined) {
-        console.log(e);
-        throw e;
+      if (retries == 0 || error.logs !== undefined) {
+        console.log(error);
+
+        throw error;
```
ananas-block, line 78: I believe this is the part that we threw the cryptic errors, and it looks unchanged to me, is this on purpose?

ananas-block (changes requested):

PR #213, opened 2023-08-03
feat: browser support

SwenSchaeferjohann:
todo
- [x] test with browser


light-system-programs/tests/verifier_tests.ts
```diff
@@ -342,6 +339,7 @@ describe("Verifier Zero and One Tests", () => {
         tmp_tx.transactionInputs.publicInputs.inputNullifier[i] = new Array(
           32,
         ).fill(2);
+        await sleep(SLEEP_BUFFER);
```
ananas-block, line 342: as discussed we could make the sleep depend on the platform.
This is how detect the platform in the cli:
https://github.com/Lightprotocol/light-protocol/blob/main/cli/src/psp-utils/download.ts#L36

light-zk.js/src/account.ts
```diff
@@ -104,7 +104,7 @@ export class Account {
       // burnerSeed can be shared since hash cannot be inverted - only share this for app utxos
       // sharing the burnerSeed saves 32 bytes in onchain data if it is require to share both
       // the encryption and private key of a utxo
-      this.burnerSeed = new BN(bs58.decode(seed)).toBuffer("be", 32);
+      this.burnerSeed = new BN(bs58.decode(seed)).toArrayLike(Buffer, "be", 32);
```
ananas-block, line 107: we should do another pr where we change this in all places where BN.toBuffer/toArray is used

relayer/src/services/relayService/addRelayJob.ts
```diff
@@ -0,0 +1,16 @@
+import { TransactionInstruction } from "@solana/web3.js";
+import { relayQueue } from "../../db/redis";
+
+export async function addRelayJob({
```
ananas-block, line 4: I deleted this file on purpose, to consolidate files and remove directories if we can,
Personally I don't like to have a lot of files with  less than 20 lines each.

relayer/src/utils/createNonce.ts
```diff
@@ -0,0 +1,3 @@
+import nacl from "tweetnacl";
+export const createNonce = (): string =>
```
ananas-block, line 2: why is this in it's own file?
(I believe this is used in one place. Thus, let's put it there.)

relayer/src/services/relayService/handleRelayRequest.ts
```diff
@@ -131,7 +115,8 @@ export async function handleRelayRequest(req: any, res: any) {
       `/handleRelayRequest - req.body.instructions: ${req.body.instructions}`,
     );
 
-    const job = await addRelayJob({ instructions });
+    const nonce = createNonce();
+    const job = await addRelayJob({ instructions, nonce });
```
ananas-block, line 119: What's the purpose of the nonce?
I think it is not a good identifier for the job. The identifier should be based on the instructions, so that in case the connection to the user drops the user can query the state of it's tx.

relayer/src/setup/index.ts
```diff
@@ -1,8 +1,81 @@
...
+} from "../utils/provider";
+import { PublicKey } from "@solana/web3.js";
+import { readFileSync, writeFileSync } from "fs";
+import { AnchorProvider, BN } from "@coral-xyz/anchor";
 
 export const testSetup = async () => {
-  const providerAnchor = await getAnchorProvider();
-  // TODO: use updated -- buildscript -> add relayer tests
-  await createTestAccounts(providerAnchor.connection);
+  const anchorProvider = await getAnchorProvider();
```
ananas-block, line 17: Are we using the relayer test setup anywhere?
(if not let's remove)

ananas-block (changes requested):

light-system-programs/tests/verifier_tests.ts
SwenSchaeferjohann: 👍 for now just importing from psp-utils so we don't have dupe code but probably makes sense to put this separately at some point? any ideas?

relayer/src/setup/index.ts
SwenSchaeferjohann: yes, renamed to relayerSetup (e.g. inits relayerRecipient when relayer starts) this was previously one source of error since for example the CLI tests start the relayer but don't execute Pre-hooks found in other tests

relayer/src/utils/createNonce.ts
SwenSchaeferjohann: done

relayer/src/services/relayService/handleRelayRequest.ts
SwenSchaeferjohann: good point, replaced with stringified instructions' data concatenated  for job.id.
is used to ID jobs in queue (should be unique)

ananas-block: what about hashing the stringified instructions data?
(now you have a job id which is really long/the actual data)

SwenSchaeferjohann:
- [ ]  move getSystem to zkjs

light-system-programs/tests/verifier_tests.ts
SwenSchaeferjohann: update: moved to util in zkjs

light-system-programs/tests/test-utils/user-utils.ts
```diff
@@ -30,6 +30,9 @@ export async function performShielding({
     const provider = await Provider.init({
       wallet: environmentConfig.providerSolanaKeypair,
       relayer: environmentConfig.relayer,
+      confirmConfig: {
+        commitment: "confirmed",
+      },
```
ananas-block, line 35: we have a confirm config constant

light-system-programs/package-lock.json
ananas-block, line 1: pls remove package-lock.json

relayer/src/services/relayService/handleRelayRequest.ts
```diff
@@ -4,10 +4,41 @@ import {
...
+  let concatenatedData = new Uint8Array([]);
+  ixs.forEach((ix) => {
+    concatenatedData = new Uint8Array([...concatenatedData, ...ix.data]);
+  });
+  const hashArray = sha3_256.create().update(concatenatedData).digest();
+  const hashString = Array.from(hashArray)
+    .map((byte) => byte.toString(16).padStart(2, "0"))
+    .join("");
+
+  return hashString;
```
ananas-block, line 23: Why don't you do:
```
let hasher = sha3_256.create();
ixs.forEach((ix) => {
   hasher.update( new Uint8Array([...concatenatedData, ...ix.data]));
  });
  return bs58.encode(hasher.digest());
```

relayer/src/setup/index.ts
```diff
@@ -1,8 +1,65 @@
...
+export async function relayerSetup() {
+  const anchorProvider = await getAnchorProvider();
+
+  await createTestAccounts(anchorProvider.connection);
+
+  let lookUpTable;
+  const path = "lookUpTable.txt";
+  try {
+    let lookUpTableRead = new PublicKey(readFileSync(path, "utf8"));
+    let lookUpTableInfoInit = await anchorProvider.connection.getAccountInfo(
```
ananas-block, line 25: we should unify this with the config from cli, but we can do that in a different pr

relayer/tests/browser_test.ts
```diff
@@ -0,0 +1,152 @@
...
+    // should expect to throw
+    expect(() => {
+      AnchorProvider.local("http://127.0.0.1:8899", {
+        commitment: "confirmed",
+      });
+    }).to.throw("Provider local is not available on browser.");
+  });
+
+  it("should init user, shield, transfer, unshield", async () => {
+    const message =
```
ananas-block, line 88: I thought we provide this abstraction in the user class. Is this necessary?
We should probably test the manually signed seed versus the one generated with the user class, right?

relayer/tests/browser_test.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (!walletMock.signMessage) throw new Error("Wallet not connected!");
+    const provider = await Provider.init({
+      relayer: RELAYER,
+      wallet: walletMock,
+      confirmConfig: { commitment: "confirmed" },
+    });
+
+    const user = await User.init({
+      provider,
+      seed: new BN(signature).toString(),
```
ananas-block, line 111: in other places we use bs58.encode

relayer/tests/browser_test.ts
```diff
@@ -0,0 +1,152 @@
...
+      amountSol: "0.25",
+      token: "SOL",
+      recipient: testRecipient.account.getPublicKey(),
+    });
+
+    await user.unshield({
+      publicAmountSol: "1.5",
+      token: "SOL",
+      recipient: new PublicKey("ErAe2LmEKgBNCSP7iA8Z6B396yB6LGCUjzuPrczJYDbz"),
+    });
```
ananas-block, line 150: we should add asserts to make sure that the tests are successful.
For example with UserAssertHelper

relayer/tests/browser_test.ts
```diff
@@ -0,0 +1,152 @@
...
+app.post("/relayTransaction", handleRelayRequest);
+app.get("/indexedTransactions", getIndexedTransactions);
+app.get("/getBuiltMerkletree", buildMerkleTree);
+
+describe("Browser tests", () => {
+  var RELAYER: Relayer;
+  const walletMock = useWallet(ADMIN_AUTH_KEYPAIR);
+  const connection = new Connection("http://127.0.0.1:8899", "confirmed");
+
+  before(async () => {
```
ananas-block, line 54: Can you explain how these tests replicate a browser env?

ananas-block (changes requested):

relayer/tests/browser_test.ts
SwenSchaeferjohann: runScript.sh runs it after initing browser env with jsdom; 
npx mocha -r ts-node/register -r jsdom-global/register -r ./setup.jsdom.ts tests/browser_test.ts --timeout 1000000 --exit;

browser_test.ts has 2 tests: 1) executing the core functionality (can expand in future) for success and 2) for failure (using a node feature like anchorprovider in browser env)

relayer/tests/browser_test.ts
SwenSchaeferjohann: not strictly necessary. added test anyway, good catch (found the internal derivation used an empty message array since the encoding didn't work properly. fixed)

ananas-block: We should probably add a user instantiation test to the user tests in general.

relayer/tests/browser_test.ts
ananas-block: ok so jsdom is initing a browser env and all tests after are executed in this env

relayer/tests/browser_test.ts
ananas-block: ah nice you  already did that

relayer/tests/browser_test.ts
SwenSchaeferjohann: yes

relayer/tests/browser_test.ts
```diff
@@ -0,0 +1,229 @@
...
+    // need to wait for balance update to fetch current utxos
+    await user.getBalance();
+    await user.transfer({
+      amountSol: testInputs.amountSol,
+      token: testInputs.token,
+      recipient: recipientAccount.getPublicKey(),
+    });
+
+    // await waitForBalanceUpdate(testStateValidator, user);
+    await sleep(6000);
```
ananas-block, line 226: What's the rationale for using sleep over waitForBalanceUpdate?
(Please remove one or the other, although I think waitForBalanceUpdate is more robust)

ananas-block (changes requested):
One last thing otherwise looks good.

General remark: The Pr got a lot bigger than 500 additions even when accounting for added tests. We should try to make Prs smaller in the future.

ananas-block (approved):

PR #215, opened 2023-08-10
feat: rock paper scissors refactor

ananas-block:
Versions:
- bump cli to alpha 18
- bump zk.js to alpha 13

Feat:
- enable PSPs to use pdas
- commented blinding check because it a usability security check which prevents our hardcoded blinding in rock paper scissors
- added checkAppData to throw an error if an appData attribute is not set
- support for multiple appUtxos
- support for additonal native in utxos
- auto generation of isAppInUtxo array

Fix:
- getVerifierProgramId was broken because anchor includes all constants in alphabetical order in the idl now
- createOutUtxos - 
- compileAndProve transactions enable controlling shuffling of utxos - (shuffling breaks psp tx right now)


light-zk.js/src/transaction/transaction.ts
```diff
@@ -820,12 +820,18 @@ export class Transaction {
       );
 
       let inputsVec = (await coder.encode(accountName, inputs)).subarray(8);
+      // TODO: check whether app account names overlap with system account names and throw an error if so
+      let appAccounts = {};
```
SwenSchaeferjohann, line 824: shouldnt this be an array?

light-zk.js/src/types/programParameters.ts
```diff
@@ -4,4 +4,5 @@ export type ProgramParameters = {
   verifierIdl: Idl;
   inputs: any; // object of proof and other inputs
   path: string;
+  accounts?: any; // object of accounts
```
SwenSchaeferjohann, line 7: why not replace `any` with the type?

light-zk.js/src/utxo.ts
```diff
@@ -132,16 +132,16 @@ export class Utxo {
...
+    //   throw new UtxoError(
+    //     UtxoErrorCode.BLINDING_EXCEEDS_SIZE,
+    //     "constructor",
 
-        `Bliding ${blinding}, exceeds size of 31 bytes/248 bit.`,
-      );
-    }
+    //     `Bliding ${blinding}, exceeds size of 31 bytes/248 bit.`,
+    //   );
+    // }
```
SwenSchaeferjohann, line 144: can this be removed?

SwenSchaeferjohann (commented):
lgtm once above questions resolved  can merge

light-zk.js/src/transaction/transaction.ts
ananas-block: Anchor requires an account object. This way we can just do:
```
tx.accounts({...appAccounts})
```

light-zk.js/src/types/programParameters.ts
ananas-block: We don't know what accounts are added in a psp before hand. The type needs to be generic. In the future we should make it depend on the idl.

light-zk.js/src/utxo.ts
ananas-block: Good point I replaced the check with a better one

light-zk.js/src/utxo.ts
```diff
@@ -131,17 +132,13 @@ export class Utxo {
...
-    //     "constructor",
-
-    //     `Bliding ${blinding}, exceeds size of 31 bytes/248 bit.`,
-    //   );
-    // }
+    if (!blinding.eq(blinding.mod(FIELD_SIZE))) {
+      throw new UtxoError(
+        UtxoErrorCode.BLINDING_EXCEEDS_FIELD_SIZE,
+        "constructor",
+        `Bliding ${blinding}, exceeds field size.`,
```
SwenSchaeferjohann, line 139: just small typo bliding

ananas-block:
```suggestion
        `Blinding ${blinding}, exceeds field size.`,
```

SwenSchaeferjohann (approved):

PR #253, opened 2023-09-14
fix: zk.js - remove chai usage & add custom utils errors

ananas-block:
Problem:
- chai is required as a dependency for any usage of zk.js

Solution:
- replace asserts in zk.js with if conditionals which throw


light-zk.js/src/test-utils/createAccounts.ts
```diff
@@ -204,12 +203,13 @@ export async function createMintWrapper({
...
       confirmConfig,
     );
-    assert(
-      (await connection.getTransaction(res, {
-        commitment: "confirmed",
-      })) != null,
-      "create mint account failed",
-    );
+    const transactionResult = await connection.getTransaction(res, {
+      commitment: "confirmed",
```
SwenSchaeferjohann, line 207: Can we use the confirmConfig for this?

light-zk.js/src/test-utils/initLookUpTable.ts
```diff
@@ -175,7 +174,9 @@ export async function initLookUpTableTest(
       lookupTableAddress,
       "confirmed",
     );
-    assert(lookupTableAccount != null);
+    if (!lookupTableAccount) {
+      throw new Error("lookupTableAccount is null or undefined");
```
SwenSchaeferjohann, line 178: Should we define and import a standardised error here?

light-zk.js/src/test-utils/testChecks.ts
SwenSchaeferjohann, line 1: Are we using these files outside tests?

light-zk.js/src/test-utils/initLookUpTable.ts
ananas-block: done, noticed that some errors were not custom in other functions either. Replaced those errors with custom ones as well.

SwenSchaeferjohann (approved):

PR #277, opened 2023-09-27
chore: add ESLint configurations

sergeytimoshin:
ESLint configurations for the prover.js, relayer, zk.js, and cli.


Shigoto-dev19:
Does it make sense to use an auto-lint/hook as in #316

vadorovsky (approved):

PR #278, opened 2023-09-27
refactor: move proof gen to account

ananas-block:
This pr is based on https://github.com/Lightprotocol/light-protocol/pull/275 please merge in sequence.

### Problem:
 - secrets are necessary to generate light system proofs

### Solution:
- move proof gen into account class and add private key class internally so that secrets are not exposed

### Changes:
- move proof gen from transaction to account
- add addPrivateKey method to account.ts -> it can only add the private key of the account. This clashes with the way we so far added empty/filling utxos.
- add account as input to TransactionParameters so that filling utxos are all owned by the same account not random accounts
- add isFillingUtxo marker to Utxo
  We don't want to decrypt a lot of random utxos therefore  if utxo is filling utxo make the prefix random, so that we don't decrypt  these utxos by default.

### DIFF
--- a/system-programs/tests/functional_tests.ts
+++ b/system-programs/tests/functional_tests.ts
@@ -235,6 +235,7 @@ describe("verifier_program", () => {
       action: Action.SHIELD,
       poseidon: POSEIDON,
       verifierIdl: verifierIdl,
+      account: KEYPAIR,
     });
     let transactionTester = new TestTransaction({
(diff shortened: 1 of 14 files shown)


vadorovsky (commented):
Just one nit for now. I will take a deeper look once ready for review.

vadorovsky (approved):

PR #282, opened 2023-09-28
chore: psp-examples + github workflow

sergeytimoshin:


.github/workflows/light-sdk-tests.yml
```diff
@@ -48,3 +48,8 @@ jobs:
             fi
             cd -
           done
+      - name: psp-examples
```
ananas-block, line 51: Psp examples ci is now being executed two times after both iterations of the sdk ci tests.
What do you think about creating another workflow which just executes psp-examples?
(Essentially you could duplicate the file and delete line 37-50.)

sergeytimoshin: Yeah, good idea, I'll do it now.

PR #288, opened 2023-09-29
refactor(zk.js): move create account with signature into account

ananas-block:
### Problem:
- when creating a user from a wallet or keypair the seed is exposed in user class

### Changes:
- move signature generation into account
- a user can be instantiated with a seed, account, or wallet of the provider
- if an account or seed is provided these will be used over the keypair of the provider
- both account and seed cannot be defined at the same time
- added the following methods to account:
  - createFromSeed
  - createFromSolanaKeypair
  - createFromBrowserWallet


zk.js/src/test-utils/functionalCircuit.ts
```diff
@@ -70,15 +71,15 @@ export async function functionalCircuitTest(
...
 
   try {
     tx.proofInput.inIndices[0][1][1] = "1";
     // TODO: investigate why this does not kill the proof
     tx.proofInput.inIndices[0][1][0] = "1";
-    await tx.getProof();
+    await tx.getProof(account);
     x = false;
   } catch (error) {
     // assert.isTrue(error.toString().includes("CheckIndices_3 line:"));
```
SwenSchaeferjohann, line 89: commented assert

SwenSchaeferjohann (commented):
overall lgtm - let's merge this. if we need to make any additional changes wrt wallet adapter compatibility we'll figure it out when testing with ui's

zk.js/src/test-utils/functionalCircuit.ts
ananas-block: Good catch, the test would have failed if it didn't throw an error. The commented account was just the precise part of the check. 
I have added an equivalent check now.

PR #290, opened 2023-09-29
feat(relayer): added getRelayerInfo

ananas-block:
Problem:
- relayerPubkey and relayerRecipientSol need to be configured manually
- inconsistencies cause transactions to fail

Changes:
- relayer server:
  - add getRelayerInfo path to relayer server (returns current relayerPubkey, relayerRecipientSol, relayerFee, relayerFeeHigh)
- Relayer class zk.js
  - added Relayer.initFromUrl and relayer.syncRelayerInfo(url: string) which does a request to getRelayerInfo()
- CLI:
    - added config fields for relayerFee and highRelayerFee
    - added syncRelayer flag to config command which syncs the relayer info

### DIFF
--- a/cli/src/utils/utils.ts
+++ b/cli/src/utils/utils.ts
@@ -120,12 +120,13 @@ export const getRelayer = async (localTestRelayer?: boolean) => {
       });
       return relayer;
     } else {
+      const config = getConfig();
       relayer = new Relayer(
-        getRelayerPublicKey(),
-        getRelayerRecipient(),
-        RELAYER_FEE,
-        TOKEN_ACCOUNT_FEE,
-        getRelayerUrl()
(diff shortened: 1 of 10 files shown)


cli/src/utils/utils.ts
```diff
@@ -215,6 +218,7 @@ export const setPayer = (key: string) => {
 };
 import { existsSync } from "fs";
 import { config } from "dotenv";
+import { compileFunction } from "vm";
```
SwenSchaeferjohann, line 221: where is it used?

ananas-block: it's not must be an auto import good catch

ananas-block: fixed

vadorovsky (approved):

PR #296, opened 2023-10-03
fix: pass provider explicitly in program instantiation

SwenSchaeferjohann:
problem
- if the `provider` instance isn't explicitly passed to the anchor Program constructor Anchor automatically instantiates a dummy provider with `Provider.local()` - which always uses localhost irrespective of the ANCHOR_PROVIDER_URL environment variable.

fix
- always explicitly pass the anchorprovider


vadorovsky (approved):
Looks good, thanks!

PR #312, opened 2023-10-06
chore: cancel previously running github actions on push

sergeytimoshin:


vadorovsky (approved):
Amazing, I need to apply it in Aya :sweat_smile:

PR #314, opened 2023-10-06
perf: add benchmarks for light circuits:  masp2 & app4

Shigoto-dev19:
Changes:
- Add a new benchmarks dir inside circuit-lib/ciruit-lib.circom.
- Use [micro-bmark](https://github.com/paulmillr/micro-bmark) package for micro benchmarks.
- Benchmark `transaction_masp2` and `transaction_app4` circuits indirectly based on different merkle tree heights in the range [18, 20, 22, 24, 26].
- Add README.md containing:
  - Build & Bench script instructions
  - Circuit statistics for both circuits based on different Merkle tree heights.
  - Proving time for both circuits based on different Merkle tree heights.
- The benchmark template is refactored in a way to be concise and reusable for light transaction circuits: `async function mark_transaction_account(tree_height: number, app = false)`

### DIFF
--- a/circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
+++ b/circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
@@ -1,0 +1,23 @@
+import { BenchmarkLightCircuit, circuitParameterVariants } from "./utils";
+
+async function bench() {
+  const maspParameterVariants: circuitParameterVariants = {
+    merkleTreeHeightVariants: [18, 20, 22, 24, 26],
+    nInputUtxosVariants: [2],
+    nOutputUtxosVariants: [2],
+    outputPath: "./artifacts/bench_circuits/light",
+  };
+  const benchMasp = new BenchmarkLightCircuit(maspParameterVariants);
+  await benchMasp.benchmark(4);
+
+  const appParameterVariants: circuitParameterVariants = {
+    merkleTreeHeightVariants: [18, 20, 22, 24, 26],
+    nInputUtxosVariants: [4],
+    nOutputUtxosVariants: [4],
+    outputPath: "./artifacts/bench_circuits/light",
+  };
+  const benchApp = new BenchmarkLightCircuit(appParameterVariants, true);
+  await benchApp.benchmark(4);
+}
+
+bench();
--- a/circuit-lib/circuit-lib.circom/benchmarks/light/utils.ts
+++ b/circuit-lib/circuit-lib.circom/benchmarks/light/utils.ts
@@ -1,0 +1,335 @@
+const bench = require("micro-bmark");
+const { compare, run } = bench;
+const snarkjs = require("snarkjs");
+const buildPoseidonOpt = require("circomlibjs").buildPoseidonOpt;
+
+import { genRandomSalt as generateRandomFieldElement } from "maci-crypto";
+import { execSync } from "child_process";
+import { resolve } from "path";
+import { writeFileSync, existsSync, promises } from "fs";
+import { Keypair as SolanaKeypair, PublicKey } from "@solana/web3.js";
+import { BN } from "@coral-xyz/anchor";
+
+import {
+  Account,
+  Utxo,
+  FEE_ASSET,
+  Provider as LightProvider,
+  MINT,
+  Transaction,
+  TransactionParameters,
+  Action,
+  IDL_VERIFIER_PROGRAM_ZERO,
+  IDL_VERIFIER_PROGRAM_TWO,
+  STANDARD_SHIELDED_PRIVATE_KEY,
+  STANDARD_SHIELDED_PUBLIC_KEY,
+} from "@lightprotocol/zk.js";
+
+import { bs58 } from "@coral-xyz/anchor/dist/cjs/utils/bytes";
+import { MerkleTree } from "../../../circuit-lib.js/lib/index.js";
+import { stringifyBigInts } from "maci-crypto";
+
+// This will extract the type of the constructor parameters of Transaction CLass
+type TransactionInputs = ConstructorParameters<typeof Transaction>[0];
+
+export type circuitParameterVariants = {
+  merkleTreeHeightVariants: number[];
+  nInputUtxosVariants: number[];
+  nOutputUtxosVariants: number[];
+  outputPath: string;
+};
+
+type CircuitModel = {
+  name: string;
+  merkleTreeHeight: number;
+  merkleTree?: MerkleTree;
+  nInputUtxos: number;
+  nOutputUtxos: number;
+};
+
+// skip unwanted logs
+console.log = (function (originalLog) {
+  const skipPatterns = [
+    "Provider is not defined.",
+    "Unable to fetch rootIndex. Setting root index to 0 as a default value.",
+    // Add more patterns if needed
+  ];
+
+  return function (...args) {
+    const message = args.join(" ");
+
+    if (!skipPatterns.some((pattern) => message.includes(pattern))) {
+      originalLog.apply(console, args);
+    }
+  };
+})(console.log);
+
+export class BenchmarkLightCircuit {
+  app: boolean;
+  circuitType: string;
+  overwrite: boolean;
+  circuitParameterVariants: circuitParameterVariants;
+  circuitModels: Array<CircuitModel>;
+  merkleTreeElements: Array<string>;
+  poseidon: any;
+
+  constructor(
+    circuitParameterVariants: circuitParameterVariants,
+    app = false,
+    overwrite = true,
+  ) {
+    this.app = app;
+    this.circuitType = app ? "app" : "masp";
+    this.overwrite = overwrite;
+    this.circuitParameterVariants = circuitParameterVariants;
+    this.circuitModels = [];
+    this.merkleTreeElements = Array.from({ length: 2 ** 13 }, () =>
+      generateRandomFieldElement().toString(),
+    );
+  }
+
+  private async ensureDirectoryExists(dirPath: string) {
+    try {
+      await promises.access(dirPath);
+    } catch (error) {
+      if (error.code === "ENOENT") {
+        // If directory doesn't exist, create it
+        await promises.mkdir(dirPath, { recursive: true });
+      } else {
+        // Rethrow the error if it's not the expected "directory doesn't exist" error
+        throw error;
+      }
+    }
+  }
+
+  private combineArrays(...arrays) {
+    if (arrays.length === 0) return [[]];
+    const [first, ...rest] = arrays;
+    const suffixes = this.combineArrays(...rest);
+    return first.flatMap((item) => suffixes.map((suffix) => [item, ...suffix]));
+  }
+
+  private async generateCircomMainFile(
+    merkleTreeHeight: number,
+    nInputUtxos: number,
+    nOutputUtxos: number,
+  ) {
+    const circomTemplate =
+      `
+    pragma circom 2.0.0;
+    include "../../../src/light/transaction_${this.circuitType}.circom";
+
+    // 2 in 2 out 3 assets (min to do a swap)
+    component main {
+        public [
+            root,
+            inputNullifier,
+            outputCommitment,
+            publicAmountSpl,
+            txIntegrityHash,
+            publicAmountSol,
+            publicMintPubkey` +
+      `${this.app ? "," : ""}
+            ${this.app ? "publicAppVerifier," : ""}
+            ${this.app ? "transactionHash" : ""}
+            `.trimEnd() +
+      `\n        ]
+    } = TransactionAccount(
+        ${merkleTreeHeight},
+        ${nInputUtxos},
+        ${nOutputUtxos},
+        184598798020101492503359154328231866914977581098629757339001774613643340069,
+        0,
+        1,
+        3,
+        2,
+        2
+    );
+        `;
+    const circuitName = `transaction${nInputUtxos}in${nOutputUtxos}out${
+      this.app ? "App" : "Masp"
+    }MTH${merkleTreeHeight}`;
+
+    this.circuitModels.push({
+      name: circuitName,
+      merkleTreeHeight,
+      nInputUtxos,
+      nOutputUtxos,
+    });
+    const finalPath =
+      this.circuitParameterVariants.outputPath + "/" + circuitName + ".circom";
+
+    // Check if file already exists
+    if (!this.overwrite && existsSync(finalPath)) {
+      return;
+    }
+
+    // Write the generated content to the outputPath
+    writeFileSync(finalPath, circomTemplate.trim());
+  }
+
+  async generateCircomMainFiles() {
+    const {
+      merkleTreeHeightVariants,
+      nInputUtxosVariants,
+      nOutputUtxosVariants,
+    } = this.circuitParameterVariants;
+
+    await this.ensureDirectoryExists(this.circuitParameterVariants.outputPath);
+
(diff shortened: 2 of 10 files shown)


circuit-lib/circuit-lib.circom/tests/light/transactionApp/transactionApp4Main18.circom
ananas-block, line 1: I think these files can be in gitignore?
For example just put them into artifacts?

At benchmark time you can just autogenerate the files for example like this:

```
import { writeFileSync, existsSync } from 'fs';

function generateCircomFile(
    merkleTreeHeight: number,
    nInputUtxos: number,
    nOutputUtxos: number,
    outputPath: string
) {
    const circomTemplate = `
pragma circom 2.0.0;
include "../../../src/light/transaction_app.circom";
// 2 in 2 out 3 assets (min to do a swap)
component main {
    public [
        root,
        inputNullifier,
        outputCommitment,
        publicAmountSpl,
        txIntegrityHash,
        publicAmountSol,
        publicMintPubkey,
        publicAppVerifier,
        transactionHash
    ]
} = TransactionAccount(
    ${merkleTreeHeight},
    ${nInputUtxos},
    ${nOutputUtxos},
    184598798020101492503359154328231866914977581098629757339001774613643340069,
    0,
    1,
    3,
    2,
    2
);
    `;

    // Check if file already exists
    if (existsSync(outputPath)) {
        console.log(`File ${outputPath} already exists. Generation aborted to prevent overwriting.`);
        return;
    }

    // Write the generated content to the outputPath
    writeFileSync(outputPath, circomTemplate.trim());
}

// Example usage:
generateCircomFile(18, 4, 4, 'path_to_generated/main.circom');
```

circuit-lib/circuit-lib.circom/scripts/buildPtau.sh
```diff
@@ -0,0 +1,57 @@
...
+# Create a directory for the ptau file
+mkdir -p ./artifacts/ptau 
+
+POWERS_OF_TAU=19 # circuit will support max 2^POWERS_OF_TAU constraints
+if [ ! -f ./ptau$POWERS_OF_TAU ]; then
+  echo "Downloading powers of tau file"
+  curl -L https://hermez.s3-eu-west-1.amazonaws.com/powersOfTau28_hez_final_$POWERS_OF_TAU.ptau --create-dirs -o ./artifacts/ptau/pot$POWERS_OF_TAU.ptau
+fi
+
+
```
ananas-block, line 26: lot's of empty lines here, pls remove

circuit-lib/circuit-lib.circom/scripts/buildTestCircuit.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# # Ceremony just like before but for zkey this time
+pnpm snarkjs zkey contribute ./artifacts/$1/$1_0000.zkey ./artifacts/$1/$1_0001.zkey \
+    --name="First $1 contribution" -v -e="$(head -n 4096 /dev/urandom | openssl sha1)"
+
+pnpm snarkjs zkey contribute ./artifacts/$1/$1_0001.zkey ./artifacts/$1/$1_0002.zkey \
+    --name="Second $1 contribution" -v -e="$(head -n 4096 /dev/urandom | openssl sha1)"
+
+pnpm snarkjs zkey contribute ./artifacts/$1/$1_0002.zkey ./artifacts/$1/$1_0003.zkey \
+    --name="Third $1 contribution" -v -e="$(head -n 4096 /dev/urandom | openssl sha1)"
```
ananas-block, line 36: There is no need to do more than one contribution for testing. Please remove the additional contributions to reduce compile time.

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
```diff
@@ -0,0 +1,169 @@
...
+  return function (...args) {
+    const message = args.join(" ");
+
+    if (!skipPatterns.some((pattern) => message.includes(pattern))) {
+      originalLog.apply(console, args);
+    }
+  };
+})(console.log);
+
+const iterations = 100;
```
ananas-block, line 139: you could check whether circuits are built here.
If not create circuit files and build the circuits here since it doesn't count into benchtime correct?

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
```diff
@@ -0,0 +1,169 @@
...
+  const tx = new Transaction(txInputs);
+  await tx.compile(lightProvider.poseidon, account);
+
+  addPrivateKey(tx, account);
+  if (app) delete tx.proofInput.inPublicKey;
+
+  return tx.proofInput;
+}
+
+async function mark_transaction_account(tree_height: number, app = false) {
```
ananas-block, line 104: why mark...?
if you want to keep it short feel free to use bench...

ananas-block: Additionally you could make this function a generic utility for any circuit benchmark by refactoring it to sth like this:
```
async function generateProof(
  name: string,
  generateInputsFunction: (...args: any[]) => Promise<any>,
  ...inputFunctionArgs: any[]
) {
  const wasm_path = `./artifacts/${name}/${name}.wasm`;
  const zkey_path = `./artifacts/${name}/${name}.zkey`;

  const proofInputs = await generateInputsFunction(...inputFunctionArgs);
  const res = await snarkjs.groth16.fullProve(
    stringifyBigInts(proofInputs),
    wasm_path,
    zkey_path
  );

  return res;
}
```

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
```diff
@@ -0,0 +1,169 @@
...
+import { stringifyBigInts } from "maci-crypto";
+
+// This will extract the type of the constructor parameters of Transaction CLass
+type TransactionInputs = ConstructorParameters<typeof Transaction>[0];
+
+function addPrivateKey(transaction: Transaction, account: Account) {
+  transaction.proofInput["inPrivateKey"] = transaction.params.inputUtxos.map(
+    (utxo: Utxo) => {
+      if (utxo.publicKey == account.pubkey) {
+        return account.privkey;
```
ananas-block, line 32: This is soon not going to work anymore because we will make account.privkey private.
Please find a different solution.
For example you could do a mock key pair which uses a random 31 bytes number as private key and the poseidon hash of it as pubkey.

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
ananas-block: We should be able to execute benchmarks including building circuits in one command :)

ananas-block (changes requested):
Thanks, looks good in general a couple comments to make it more general and keep the repo easier to maintain.

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
```diff
@@ -0,0 +1,169 @@
...
+    publicKey: account.pubkey,
+    assetLookupTable: lightProvider.lookUpTables.assetLookupTable,
+    verifierProgramLookupTable:
+      lightProvider.lookUpTables.verifierProgramLookupTable,
+  });
+
+  const mockPubkey: PublicKey = SolanaKeypair.generate().publicKey;
+
+  let txParams = new TransactionParameters({
+    outputUtxos: [deposit_utxo1],
```
ananas-block, line 62: I am surprised that the bench results show a much lower increase in proving time than the results of the merkle proof benches. Can you add as many in utxos as the circuit allows to make sure we get a comparable number?

Merkle proofs are just enforced when amounts of input utxos are non zero and I think for empty utxos we just use zeros as merkle proofs. Thus the benchmarks might not be representative.

circuit-lib/circuit-lib.circom/tests/light/transactionApp/transactionApp4Main18.circom
Shigoto-dev19: the thought came to my mind, but I wasn't sure if the light API is flexible enough to generate the circuit inputs for each variant.

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
Shigoto-dev19: mark is a notation to track a function's performance in general. I wrote following the benchmark framework as well. Anyway, I changed into `bench` for better readability.

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
Shigoto-dev19: i'll generate the same keypair by the static methods:
```typescript
  const account_privateKey = Account.generateShieldedPrivateKey(seed32, poseidon);
  const account_publicKey = Account.generateShieldedPublicKey(account_privateKey, poseidon);
```
but if you want `privKey` to be private then you also need to make `seed` in the account class private as well.

circuit-lib/circuit-lib.circom/tests/light/transactionApp/transactionApp4Main18.circom
Shigoto-dev19: thanks for the snippet!

circuit-lib/circuit-lib.circom/benchmarks/index.ts
```diff
@@ -0,0 +1,319 @@
+const bench = require("micro-bmark");
+const { compare, run } = bench;
+const snarkjs = require("snarkjs");
```
ananas-block, line 3: This is an unexpected location for this code.
Please move to a utils file in the light directory and ideally find a better name for it.
(I think the light directory is the right place for this logic because it contains a lot of code which will only be used for light protocol circuit benchmarks.)

circuit-lib/circuit-lib.circom/benchmarks/light/bench.ts
ananas-block, line 1: I liked the prior name, because it was the name of the circuit file right?
I think to name bench files the same as the circuit makes the bench easy to find.

circuit-lib/circuit-lib.circom/benchmarks/index.ts
```diff
@@ -0,0 +1,319 @@
...
+    if (!skipPatterns.some((pattern) => message.includes(pattern))) {
+      originalLog.apply(console, args);
+    }
+  };
+})(console.log);
+
+export class BenchmarkLightCircuit {
+  app: boolean;
+  circuitType: string;
+  rewrite: boolean;
```
ananas-block, line 68: what does rewrite mean?

circuit-lib/circuit-lib.circom/benchmarks/index.ts
```diff
@@ -0,0 +1,319 @@
...
+    // Check if file already exists
+    if (!this.rewrite && existsSync(finalPath)) {
+      return;
+    }
+
+    // Write the generated content to the outputPath
+    writeFileSync(finalPath, circomTemplate.trim());
+  }
+
+  async batchGenerateCircomFileVariants() {
```
ananas-block, line 158:
```suggestion
  async generateCircomMainFiles() {
```

circuit-lib/circuit-lib.circom/benchmarks/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  private combineArrays(...arrays) {
+    if (arrays.length === 0) return [[]];
+    const [first, ...rest] = arrays;
+    const suffixes = this.combineArrays(...rest);
+    return first.flatMap((item) => suffixes.map((suffix) => [item, ...suffix]));
+  }
+
+  private generateCircomFile(
```
ananas-block, line 101:
```suggestion
  private generateCircomMainFile(
```

circuit-lib/circuit-lib.circom/benchmarks/index.ts
ananas-block: probably regenerate is more applicable?

ananas-block: or overwrite

circuit-lib/circuit-lib.circom/benchmarks/index.ts
```diff
@@ -0,0 +1,319 @@
...
+      publicKey: account.pubkey,
+      assetLookupTable: lightProvider.lookUpTables.assetLookupTable,
+      verifierProgramLookupTable:
+        lightProvider.lookUpTables.verifierProgramLookupTable,
+    });
+
+    const mockPubkey: PublicKey = SolanaKeypair.generate().publicKey;
+
+    let txParams = new TransactionParameters({
+      outputUtxos: [deposit_utxo1],
```
ananas-block, line 226: You didn't add in utxos yet, so you are still hashing zeros but you probably know that.

ananas-block (changes requested):

circuit-lib/circuit-lib.circom/benchmarks/index.ts
Shigoto-dev19: renamed it as utils and moved it inside ./benchmarks/light

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
Shigoto-dev19: done!

circuit-lib/circuit-lib.circom/benchmarks/index.ts
Shigoto-dev19: I am going for overwrite

circuit-lib/circuit-lib.circom/benchmarks/index.ts
Shigoto-dev19: yes, that's why I didn't resolve it from the earlier requested changes.

Shigoto-dev19: done!

circuit-lib/circuit-lib.circom/benchmarks/light/bench.ts
Shigoto-dev19: renamed it as `transaction_light.ts` => the old name

circuit-lib/circuit-lib.circom/benchmarks/light/transaction_light.ts
Shigoto-dev19: added 2^13 elements => random Field Elements that simulate hashes.

circuit-lib/circuit-lib.circom/benchmarks/light/utils.ts
```diff
@@ -0,0 +1,335 @@
...
+      publicKey: account.pubkey,
+      assetLookupTable: lightProvider.lookUpTables.assetLookupTable,
+      verifierProgramLookupTable:
+        lightProvider.lookUpTables.verifierProgramLookupTable,
+    });
+
+    const mockPubkey: PublicKey = SolanaKeypair.generate().publicKey;
+
+    const txParams = new TransactionParameters({
+      outputUtxos: [deposit_utxo1],
```
ananas-block, line 238: The input utxos are still missing, thus the benchmarks number are probably not representative.
Please add as many input utxos as the respective circuit allows.
https://github.com/Lightprotocol/light-protocol/blob/main/zk.js/tests/transaction.test.ts#L244

ananas-block (changes requested):

PR #317, opened 2023-10-07
feat: hosted relayer docker deploy script

SwenSchaeferjohann:
- various devnet-related fixes for hosted relayer deployment
including
- bug fix: closeMerkleTreeUpdateState (added test, changed program)


zk.js/tests/system-programs/merkle_tree_tests.ts
```diff
@@ -47,6 +47,7 @@ import {
   BN_2,
 } from "../../src";
 import { SPL_NOOP_ADDRESS } from "@solana/spl-account-compression";
+import { closeMerkleTreeUpdateState } from "../../lib";
```
ananas-block, line 50: why do you import from lib not src?

zk.js/src/test-utils/updateMerkleTree.ts
```diff
@@ -47,7 +48,18 @@ export async function updateMerkleTreeForTest(payer: Keypair, url: string) {
       transactionMerkleTree: transactionMerkleTreePda,
     });
   } catch (err) {
+    // TODO: Revisit recovery.
+    // Rn, we're just blanked-closing the update state account on failure which
+    // might not be desirable in some cases.
     console.error("failed at updateMerkleTreeForTest", err);
+    try {
+      console.log("closing update state account...");
+      await closeMerkleTreeUpdateState(merkleTreeProgram, payer, connection);
```
ananas-block, line 57: what do you think about introducing another check in case the merkle tree update state creation fails?
(In catch check whether state already exists if yes close and retry.)

zk.js/src/merkleTree/merkleTreeUpdater.ts
```diff
@@ -187,14 +222,15 @@ const checkComputeInstructionsCompleted = async (
   return accountInfo.currentInstructionIndex.toNumber() === 56;
 };
 
+// TODO: investigate
```
ananas-block, line 225: what do you want to investigate?

relayer/src/config.ts
```diff
@@ -16,7 +16,7 @@ export function getLookUpTableVar() {
 export function setLookUpTableVar(value: string) {
   _LOOK_UP_TABLE = value;
 }
-
+export const ATTEMPTS = 1;
```
ananas-block, line 19: attempts for what?
```suggestion
export const REDIS_CONNECT_RETRIES = 1;
```

ananas-block (changes requested):
Couple small things.

zk.js/src/merkleTree/merkleTreeUpdater.ts
SwenSchaeferjohann: while debugging with prints I'd seen some stuff that looked confusing mostly about the call flow of the batchupdate, but it works - removed the comment and added to non-prio'd todos

SwenSchaeferjohann:
thx for the review! addressed your comments @ananas-block

ananas-block (approved):

PR #320, opened 2023-10-09
chore: lint on commit

sergeytimoshin:


Shigoto-dev19 (approved):

scripts/format.sh
```diff
@@ -3,4 +3,6 @@
 set -e
 
 npx nx run-many --target=format --all
+npx nx run-many --target=lint:fix --all
```
ananas-block, line 6: do we have the command lint:fix in any package?

PR #324, opened 2023-10-11
fix: robust lookuptable for relayer tests

SwenSchaeferjohann:
- preloaded lookuptable account or test-validator for use by the relayer
this
- resolves ci on main (for some reason it failed after @ananas-block merged the last one)
also
- removes unnecessary logs

In a separate PR I will refactor the overall approach to setting the LOOKUPTABLE env to remove some redundant code, but i will need to test it with hosted relayer deployments before i do that


vadorovsky (approved):
Or nvm, this is severely blocking me so I'm just going to merge it and fix the script myself

PR #331, opened 2023-10-13
feat: docker fix (works with latest repo structure)

SwenSchaeferjohann:
- extracted from https://github.com/Lightprotocol/light-protocol/pull/327 into dedicated PR for clarity
- adds a working script for relayer deployment on digitalocean


SwenSchaeferjohann:
preferably merge directly after #327

scripts/deployRelayer.sh
```diff
@@ -0,0 +1,97 @@
...
+pnpm install
+cd ..
+
+
+# build docker image and deploy to digitalocean
+docker buildx create --name mybuilder
+docker buildx use mybuilder
+docker run --privileged --rm tonistiigi/binfmt --install all
+docker buildx build --platform linux/amd64 -t relayer-app:latest . --load
+docker tag relayer-app:latest registry.digitalocean.com/v3-relayer/relayer-app:latest
```
ananas-block, line 94: Two ideas how to test this script:

1. add a test flag which let's you skip line 94-96. This way you can build docker but not deploy it.
2. separate the script into two, the first builds docker, the second uses the first to build docker and deploys it. The test would only test the first script.

ananas-block (changes requested):
I think it makes sense to add a test, otherwise looks good, to me.

scripts/buildDocker.sh
ananas-block, line 1: What about naming these files ```buildDockerRelayer.sh```, ```deployDockerRelayer.sh```, ```testDockerRelayer.sh```?

ananas-block (approved):

PR #338, opened 2023-10-16
feat: Devnet stop backward fill

SwenSchaeferjohann:
changes:
- indexer backward filling only at startup and until the oldest tx is found, then stop to reduce rpc calls
- fix docker deploy process
- debugged hosted relayer docker instance


scripts/buildDockerRelayer.sh
```diff
@@ -51,12 +49,19 @@ cleanup() {
...
 
+    echo "Deleting NPM artifacts..."
+    rm -rf $(dirname $0)/../zk.js/package-lock.json
+    rm -rf $(dirname $0)/../relayer/package-lock.json
+    rm -f $(dirname $0)/../pnpm-lock.yaml
+
+
     echo "Rebuilding workspace..."
-    $(dirname $0)/build.sh
+    # bash $(dirname $0)/build.sh
```
ananas-block, line 64: is this commented on purpose?

ananas-block (commented):
Couple questions, and please let's not add key material to the repo.

scripts/buildDockerRelayer.sh
SwenSchaeferjohann: fixed thx

Dockerfile
```diff
@@ -5,6 +5,7 @@ WORKDIR /usr/src/app
 COPY ./relayer /usr/src/app/relayer
 
 COPY ./docker-entrypoint.sh /usr/local/bin/
+RUN chmod +x /usr/local/bin/docker-entrypoint.sh
```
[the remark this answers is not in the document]
SwenSchaeferjohann: nope not necessary, removed

ananas-block (approved):

PR #354, opened 2023-10-27
feat: web-ui in monorepo

SwenSchaeferjohann:
- adds web directory to monorepo
- first app: web-app (a ref implementation for viewing & managing shielded assets)

**features:**

- view shielded assets
- view public sol balance
- view tx history
- shield
- unshield
- transfer

- adds cypress e2e tests for these actions to the CI

**additionally**
- fixes .env file checks in relayer scripts ($.env -> .env)
- makes env contents manually sourceable (web-app needs to source relayer env for tests)

we want to detect downstream breaking changes, therefore it's part of the mono repo for now.


.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+        run: |
+          (cd web/webapp && npx cypress install)
+
+      # - name: Build relayer
+      #   run: |
+      #     (cd relayer && pnpm install && pnpm build)
+
+      # - name: Build webapp
+      #   run: |
+      #     (cd web/webapp && pnpm install && pnpm build)
```
ananas-block, line 64: leftover comments?

circuit-lib/circuit-lib.js/src/index.ts
```diff
@@ -1,3 +1,3 @@
 export * from "./merkle-tree";
-export * from "./elgamal-babyjubjub";
+// export * from "./elgamal-babyjubjub";
```
ananas-block, line 2: should be uncommented again right?

relayer/.env.devnet.example
```diff
@@ -1,11 +1,11 @@
...
+export RELAYER_RECIPIENT=
+export LOOK_UP_TABLE=
+export LOCAL_TEST_ENVIRONMENT="false"
+export REDIS_ENVIRONMENT="LOCAL"
+export NETWORK="DEVNET"
+export RPC_URL=
+# export DB_PORT="25061"
+# export HOSTNAME=
+# export PASSWORD=
+# export USERNAME="default"
```
ananas-block, line 11: leftover comments?

web/webapp/public/build-circuits/transactionApp4Main.zkey
ananas-block, line 1: Right now we are storing zkey files twice.
Why don't we copy the zkey files when building the repo?

ananas-block (changes requested):
Can you expand on what the webapp does in the pr description?
Also probably makes sense to add a readme.

Additionally, I found a couple things which can probably be improved.

web/webapp/public/build-circuits/transactionApp4Main.zkey
SwenSchaeferjohann: true, however for some reason the app is not able to access the wasm files when pasting the files on the fly (as part of a prebuild step) and debugging this is out of scope for this pr. (i confirmed the build-circuit files are properly copied for example). not a big deal for now; would like to revisit once deploying apps in production  (how to prevent blowing up the bundle size twice for production-apps)

relayer/.env.devnet.example
SwenSchaeferjohann: added an inline comment - these will need to be used if folks want to access a remote DB

circuit-lib/circuit-lib.js/src/index.ts
SwenSchaeferjohann: thx. fixed.  the fix for this included having to move precompute() in elgamal to its own file and not default export it, since its using a node only dependency (progress-bar). since its only used in the .test.ts file this is fine.

vadorovsky (commented):
Just couple of nitpicks. In general, looks great (tho I can't say much about the frontend code), I'm excited to see it in action. :D

web/webapp/public/build-circuits/transactionApp4Main.zkey
SwenSchaeferjohann: nvm - found a fix for it; solution was to use the -L tag with CP such that its able to follow pnpm symlinks

vadorovsky (approved):

circuit-lib/circuit-lib.js/src/index.ts
ananas-block: we should put it in a separate sub package

ananas-block (approved):

PR #379, opened 2023-11-27
feat: balance refactor

SwenSchaeferjohann:
adds
- getTokenDataByMint
- initTokenBalance
- updateTokenBalanceWithUtxo
- addUtxoToBalance
- spendUtxo

as well as high-level wrappers for 
- serializeTokenBalance
- deserializeTokenBalance
- serializeBalance
- deserializeBalance

and unit tests for all functions
- balance.new.tests.ts

what's missing is a proper syncBalance set of functions 
that
- don't need to re-fetch all txs
- utilizes a set of e.g. (lastSyncedBlock, userAccountCreationBlock) functions that can be manually cached and help pre-filter the txs to parse. 
- there should be a "lazyInboxCheck" which can be a long-running process and complement the actual syncBalance call

- [ ] move errors

### DIFF
--- a/zk.js/src/balance/balance.ts
+++ b/zk.js/src/balance/balance.ts
@@ -2,7 +2,7 @@ import { BN } from "@coral-xyz/anchor";
 import { Utxo } from "../utxo";
 import { PublicKey } from "@solana/web3.js";
 import { BN_0, TOKEN_REGISTRY } from "../constants";
-import { getPoseidon } from "poseidon";
+import { Provider } from "../wallet";
 /**
  * the SOL asset is always the 0th index in the UTXO
  * the SPL asset is an optional 1st index in the UTXO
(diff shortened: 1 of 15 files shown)


zk.js/src/balance/balance.ts
```diff
@@ -0,0 +1,274 @@
...
+export function getTokenDataByMint(
+  mintToFind: PublicKey,
+  tokenRegistry: Map<string, TokenData>,
+): TokenData {
+  for (const value of tokenRegistry.values()) {
+    if (value.mint.equals(mintToFind)) {
+      return value;
+    }
+  }
+  throw new Error(`Token with mint ${mintToFind} not found in token registry.`);
```
ananas-block, line 37: We have custom balance errors, please use them.

zk.js/src/balance/balance.ts
```diff
@@ -0,0 +1,274 @@
...
+  tokenData: TokenData,
+  utxos?: Utxo[],
+): TokenBalance {
+  let amount = BN_0;
+  let lamports = BN_0;
+
+  if (utxos) {
+    utxos.forEach((utxo) => {
+      if (!utxo.assets[UTXO_ASSET_SPL_INDEX].equals(tokenData.mint)) {
+        throw new Error(`UTXO does not match mint ${tokenData.mint}`);
```
ananas-block, line 60: Custom error pls.

ananas-block (changes requested):
Looks good on a high level, found a couple small things here and there.

PR #383, opened 2023-11-29
refactor: circuit-lib.js – switch to noble/blake

sergeytimoshin:


circuit-lib/circuit-lib.js/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "license": "GPL-3.0",
   "scripts": {
-    "test": "ts-mocha --resolveJsonModule ./tsconfig.json -t 100000000 tests/**/*.ts --exit",
+    "test": "ts-mocha --resolveJsonModule ./tsconfig.test.json -t 100000000 tests --recursive --extension .test.ts --exit",
```
ananas-block, line 7: Why did you add the test tsconfig?

circuit-lib/circuit-lib.js/tests/indices.test.ts
```diff
@@ -2,9 +2,23 @@ import { assert } from "chai";
...
+  
+  it("noble/blake vs blake-hash", async () => {
+    const b2params = { dkLen: 32 };
+    const input = "000";
+    const a = new Uint8Array(createBlakeHash("blake512").update(input).digest().slice(0, 32));
+    const b = blake2b.create(b2params).update(input).digest();
+
+    assert.equal(a.length, b.length);
+    assert.equal(a, b);
+  });
```
ananas-block, line 20: Please move the test above into it's own file, so that the naming and contents of the test files is consistent.

ananas-block (changes requested):
Thanks, for removing this additional dependency, looks good in general.
Please also move the blake-hash dep to devDependencies so that it is completely removed from the production build.
Also please remove the commented code :)

PR #385, opened 2023-12-03
chore: add bundler to prover.js

sergeytimoshin:


.github/actions/setup-and-build/action.yml
```diff
@@ -59,6 +59,10 @@ runs:
       shell: bash
       run: pnpm install
 
+    - name: Install Playwright Browsers
+      shell: bash
+      run: npx playwright install --with-deps
+
```
SwenSchaeferjohann, line 65: wouldn't it save time if we moved this to the light-sdk-test specifically such that it only runs in there? for reference, we're installing `cypress` similarly for the e2e-tests.yml.

SwenSchaeferjohann (commented):
lgtm once 1 comment resolved :)

PR #395, opened 2023-12-10
fix: update source path in copy-circuits.sh

sergeytimoshin:
• webapp expect existence of `public/build-circuits/` path. If we enable -R flag for `cp`, and source_file for `cp` ends in a /, the contents of the directory are copied rather than the directory itself. This is not what we expect. 
• .gitignore updated, /web/webapp/public/build-circuits included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ node_modules
 **/target/
 
 **/.pnpm-store/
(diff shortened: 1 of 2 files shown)


SwenSchaeferjohann (approved):
thank you!

PR #402, opened 2023-12-12
refactor: removed scripts of unmaintained psps

ananas-block:
### Issue
- Some psp examples (multisig, private-compressed-account, rock-paper-scissors) are not maintained and tested are excluded from ci.
- Since these psps are part of the pnpm workspace we still build them, which takes a lot of time.

### Changes:
- remove package.json scripts of unmaintained psp examples

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -850,6 +850,104 @@ importers:
         specifier: 5.3.2
         version: 5.3.2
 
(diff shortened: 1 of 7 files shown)


SwenSchaeferjohann (approved):

SwenSchaeferjohann:
keeping this branch so we can apply again once coming back to these psps

PR #413, opened 2023-12-17
refactor(zk.js): remove transaction and transactionParameters

ananas-block:
### Issue:
- Transaction and TransactionParameters was only used in tests
- bugs in transaction_merkle_tree
  - layout is not aligned after struct (we already fixed it in event_merkle_tree)
  - merkle_tree_nr was not assigned

### Changes:
- removed transaction.ts, transactionParameters.ts
- adjusted tests which depended on these classes
- commented circuit and verifier security tests for because a lot of other things will change in the sdk as well and these tests don't test the sdk but the programs and circuits. We will refactor these tests once the sdk is stable again.
- fixed the bugs of transaction_merkle_tree, needed to regenerate accounts after that
- moved provider test into functional tests and deleted provider.test.ts to avoid spinning up a validator for this short test


cli/accounts/misc/REGISTERED_POOL_PDA_SOL.json
SwenSchaeferjohann, line 1: do we know why these changed?

scripts/getAccountState.sh
```diff
@@ -1,4 +1,5 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+
```
SwenSchaeferjohann, line 2: why change to bash?

zk.js/src/merkleTree/merkleTreeConfig.ts
```diff
@@ -208,7 +235,7 @@ export class MerkleTreeConfig {
       await this.getTransactionMerkleTreeAccountInfo(
         transactionMerkleTreePubkey,
       );
-    return transactionMerkleTreeAccountInfo.newest == 1;
+    return transactionMerkleTreeAccountInfo.newest.toString() == "1";
```
SwenSchaeferjohann, line 238:
```suggestion
    return transactionMerkleTreeAccountInfo.newest.toString() === "1";
```

zk.js/tests/system-programs/functional_tests.ts
```diff
@@ -88,6 +91,48 @@ describe("verifier_program", () => {
     });
   });
 
+  it("Provider", async () => {
```
SwenSchaeferjohann, line 94: nitpick: "provider should init with ..?" for clarity on what test does

zk.js/tests/system-programs/verifier_tests.ts
SwenSchaeferjohann, line 1: will disabling these tests potentially cause downstream issues?

SwenSchaeferjohann (commented):
lgtm! comments mostly nits. but i'm pretty sure using bash over sh causes issues on other machines - what's the reasoning for switching it there?
edit: nvm i think bash is fine

zk.js/src/merkleTree/merkleTreeConfig.ts
SwenSchaeferjohann: while not sure if type coercion happens in this case it's safer to do strict comparison

SwenSchaeferjohann (approved):
lgtm -- approving to unblock - pls review my comments before tho

PR #416, opened 2023-12-19
refactor(zk.js): add utxo type

ananas-block:
### Issues:
- classes are a pain to use in web uis

### Changes:
- added outUtxo type
   - this type is used for out utxos - utxo which have not been inserted into the Merkle tree yet
- added Utxo type
  - this type is used for in utxos and utxos in the balance
  - these utxos are inserted in the Merkle tree
  - utxos will usually be created from decrypted out utxos

### Notes:
- this is the first pr see sprint doc
- in subsequent prs the legacy utxo class will be replaced and removed
- for now the legacy utxo class is kept unchanged apart from renaming the anchor Utxo account to UtxoLegacy
- not sure yet whether I should implement serialization for the Utxo type as well  since we don't need it for the sdk usage flow right now see commented code


SwenSchaeferjohann (approved):
lgtm

PR #425, opened 2023-12-22
feat: refactor sendSolanaTransactions

SwenSchaeferjohann:
- relayer /relayTransaction interface:
accepts instructions, fetches blockhash, creates versionedTransactions, signs, serializes them and adds to job queue
returns signatures+blockhashinfo early. independently the worker picks up and processes the relay job.
this allows the client to confirm transactions using regular web3.js methods

- sendAndConfirmTransaction + sendAndConfirmShieldedTransaction => refactored and more concise naming i.e.
createSolanaTransactions, provider.sendAndConfirmSolanaInstructions, relayer.sendAndConfirmSolanaInstructions, and so on. 
the design allows for maximum control on the lowest level (web3js methods for signing, sending, and confirming) and respective convenience wrappers provider and provider.relayer namespacing for clarity. this makes it easier to navigate in scenarios where e.g. a client might want to execute an arbitrary shielded transaction by itself (w/o) relayer.

- added inline comments 
- enforced a placeholder return type (ActionResponseMulti) to key user actions for better typesafety. this will be made obsolete in a future pr that deals with unifying the return interface

### DIFF
--- a/zk.js/src/relayer.ts
+++ b/zk.js/src/relayer.ts
@@ -1,11 +1,15 @@
 import {
+  ConfirmOptions,
   Connection,
   PublicKey,
   RpcResponseAndContext,
   SignatureResult,
+  TransactionConfirmationStrategy,
+  TransactionInstruction,
+  TransactionSignature,
 } from "@solana/web3.js";
 import { BN } from "@coral-xyz/anchor";
-import axios from "axios";
+import axios, { AxiosResponse } from "axios";
 import {
   RelayerError,
   RelayerErrorCode,
@@ -16,6 +20,9 @@ import {
   BN_0,
   RpcIndexedTransaction,
   RelayerIndexedTransaction,
+  PrioritizationFee,
+  RelayerRelayPayload,
+  SignaturesWithBlockhashInfo,
 } from "./index";
 
 export type RelayerSendTransactionsResponse =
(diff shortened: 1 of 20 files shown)


zk.js/src/relayer.ts
```diff
@@ -83,15 +90,52 @@ export class Relayer {
...
-  ): Promise<RelayerSendTransactionsResponse> {
+  /**
+   * Convenience function for sending and confirming instructions via Light RPC node.
+   * Routes instructions to Light RPC node and confirms the last transaction signature.
+   */
+  async sendAndConfirmSolanaInstructions(
+    ixs: TransactionInstruction[],
+    connection: Connection,
+    confirmOptions?: ConfirmOptions,
+    prioritizationFee?: PrioritizationFee,
```
ananas-block, line 101: We should add the ``prioritizationFee`` to the relayer fee so that the relayer gets reimbursed for the cost.

ananas-block: We could add the prioritization fee to the relayer as a property and add it when using getRelayerFee() and make the fee variable private to ensure that getRelayerFee() is being used everywhere.

ananas-block (changes requested):
Looks good in general!
Found one thing see comments.

SwenSchaeferjohann:
closing in favor of #448

PR #430, opened 2024-01-05
refactor: rename hasher

sergeytimoshin:


ananas-block (approved):

ananas-block (changes requested):
Thanks for the renaming!
A bit confused that this pr contains the nacl pr.
Saw that native nacl (de)encryption errors are not propagated see https://github.com/Lightprotocol/light-protocol/pull/429#discussion_r1441937647.

ananas-block (approved):

PR #435, opened 2024-01-09
feat: add proof compression

ananas-block:
This pr is ready for review.
However, I want to do a new release with groth16-solana before merging because right now the groth16 solana version is tied to a commit in a branch.

Changes:
- verifier sdk:
  - added ProofCompressed for serialized compressed G1 and G2 elements
  - remove negation code, since we now negate in the client
- prover-js
  - added compression and bitmask
- programs & psps:
  - adjusted for compressed proof length

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2253,9 +2253,12 @@ dependencies = [
 [[package]]
 name = "groth16-solana"
 version = "0.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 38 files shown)


vadorovsky (approved):
Looks good, I have no remarks. I guess the only thing left to be done is pointing to the commit on main branch of groth16-solana?

ananas-block:
> Looks good, I have no remarks. I guess the only thing left to be done is pointing to the commit on main branch of groth16-solana?

exactly

PR #443, opened 2024-01-15
chore: bring back private-compressed-account

SwenSchaeferjohann:
cursor went ballistic and we dont run tests for the private-compressed-account PSP example, and I didn't catch it either.  we should consider kicking those untested ones out and into separate repos for until we're done with this wave of breaking changes.

### DIFF
--- a/psp-examples/private-compressed-account/tests/private-compressed-account.ts
+++ b/psp-examples/private-compressed-account/tests/private-compressed-account.ts
@@ -4,516 +4,30 @@
  */
 //@ts-nocheck
 import * as anchor from "@coral-xyz/anchor";
-import { BN } from "@coral-xyz/anchor";
 import { assert } from "chai";
 import {
-  Account,
-  Action,
-  airdropSol,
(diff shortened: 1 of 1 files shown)


psp-examples/private-compressed-account/tests/private-compressed-account.ts
```diff
@@ -4,606 +4,143 @@
...
+      rpcRecipientSol: wallet.publicKey,
+      rpcFee: new BN(100000),
+      payer: wallet,
+    });
+
     // The light provider is a connection and wallet abstraction.
-    // The wallet is used to derive the seed for your compressedion keypair with a signature.
-    let lightProvider = await LightProvider.init({
+    // The wallet is used to derive the seed for your compression keypair with a signature.
+    var lightProvider = await LightProvider.init({
```
sergeytimoshin, line 103: Please don't use `var`, use `let` or `const` instead.

sergeytimoshin (approved):

PR #451, opened 2024-01-29
chore: swen/refactor utxo

SwenSchaeferjohann:
refactors the state primitive
- utxo/program-utxo as explicit types
- utxo handlers closer to pure functions  / stronger typing for internal fns


zk.js/src/account.ts
```diff
@@ -418,14 +419,19 @@ export class Account {
...
-        return STANDARD_COMPRESSION_PRIVATE_KEY;
-      }
-    });
+  private addPrivateKey(
+    proofInput: any,
+    inputUtxos: (Utxo | ProgramUtxo<PlaceHolderTData>)[],
+  ) {
+    proofInput["inPrivateKey"] = inputUtxos.map(
+      (utxo: Utxo | ProgramUtxo<PlaceHolderTData>) => {
+        if ("data" in utxo) {
```
ananas-block, line 428: Why change this?
We are using the standard key pair also for filling utxos which do not have data.

ananas-block: Actually I think a better way is two ifs (not sure whether the code below works)
```suggestion
        if (STANDARD_COMPRESSION_PUBLIC_KEY === utxo.owner) { 
```

ananas-block: else should throw an error

zk.js/src/build-balance.ts
```diff
@@ -38,9 +38,9 @@ export class TokenUtxoBalance {
   tokenData: TokenData;
   totalBalanceSpl: BN;
   totalBalanceSol: BN;
-  utxos: Map<string, Utxo>; // commitment hash as key
-  committedUtxos: Map<string, Utxo>; // utxos which are
-  spentUtxos: Map<string, Utxo>; // ordered for slot spent - maybe this should just be an UserIndexedTransaction
+  utxos: Map<string, Utxo | ProgramUtxo<PlaceHolderTData>>; // commitment hash as key
+  committedUtxos: Map<string, Utxo | ProgramUtxo<PlaceHolderTData>>; // utxos which are
```
ananas-block, line 42: no need for comitted utxos

zk.js/src/build-balance.ts
```diff
@@ -212,7 +216,7 @@ export async function decryptAddUtxoToBalance({
 
   // null if utxo did not decrypt -> return nothing and continue
   if (!decryptedUtxo.value || decryptedUtxo.error) return;
-  if (decryptedUtxo.value && decryptedUtxo.value.utxoDataHash !== "0") return;
+  // if (decryptedUtxo.value && decryptedUtxo.value.dataHash !== "0") return;
```
ananas-block, line 219: commented on purpose?

ananas-block: I believe the reason for this check was to ensure that we don't decrypt program utxos and add them twice to the balance, once as normal utxo and once as program utxo.

zk.js/src/utxo/bn31.ts
```diff
@@ -0,0 +1,38 @@
+import { BN } from "@coral-xyz/anchor";
+import { bs58 } from "@coral-xyz/anchor/dist/cjs/utils/bytes";
+
+/** BN with 254-bit max size */
```
ananas-block, line 4: 248 since 31 bytes

zk.js/src/utxo/program-utxo.ts
```diff
@@ -17,75 +15,128 @@ import {
...
+  "owner"
+> & {
+  /** Public key of program that owns the utxo */
+  owner: PublicKey;
+  /** Data assigned to the utxo */
+  data: TData;
+  /** Hash of 'data' */
+  dataHash: BN31;
+  /** psp idl */
+  ownerIdl: Idl; /// TODO: remove from utxo (waste of space)
```
ananas-block, line 46: agreed in a future pr we could just get it with a function from a mapping with all system psp idls or pass it independently when needed.

zk.js/src/utxo/program-utxo.ts
```diff
@@ -464,23 +577,28 @@ export async function decryptProgramUtxo({
...
-      STANDARD_COMPRESSION_PRIVATE_KEY.toArray(),
-    );
-    const bs5832 = bs58.encode(new Uint8Array(32).fill(1));
-    account = Account.fromPrivkey(lightWasm, bs58Standard, bs5832, bs5832);
+    return decryptedProgramOutUtxo as Result<
+      ProgramUtxo<PlaceHolderTData> | null,
+      UtxoError
+    >;
   }
+  /// TODO: double check if this is needed
```
ananas-block, line 591: do we have a test that if we encrypt a utxo which has the standard public key as owner we decrypt it with the standard public key?

ananas-block: I believe we don't decrypt the public key but infer it assuming it is the public key of the account which decrypts the utxo

zk.js/src/utxo/select-in-utxos.ts
```diff
@@ -132,9 +137,9 @@ export function selectInUtxos({
   publicAmountSpl?: BN;
   publicAmountSol?: BN;
   rpcFee?: BN;
-  utxos?: Utxo[];
-  inUtxos?: Utxo[];
-  outUtxos?: OutUtxo[];
+  utxos?: (Utxo | ProgramUtxo<PlaceHolderTData>)[];
+  inUtxos?: (Utxo | ProgramUtxo<PlaceHolderTData>)[];
+  outUtxos?: (OutUtxo | ProgramOutUtxo<PlaceHolderTData>)[];
```
ananas-block, line 142: I am not sure this function makes sense for program utxos.

zk.js/src/utxo/select-in-utxos.ts
```diff
@@ -212,8 +217,9 @@ export function selectInUtxos({
   }
 
   // if mint is provided filter for only utxos that contain the mint
-  let filteredUtxos: Utxo[];
+  let filteredUtxos;
   let sumOutSpl = publicAmountSpl ? publicAmountSpl : BN_0;
+  /// TODO: review why we're using getUtxoArrayAmount which seems to be for inUtxos
```
ananas-block, line 222: it's a utility function works for both in and out utxos equally since amount arrays are the same.

zk.js/src/build-balance.ts
SwenSchaeferjohann: should not be needed anymore since program-utxo and utxo get encoded differently and programutxos should be able to go through as regular utxos at decryption/decoding i think

zk.js/src/utxo/program-utxo.ts
SwenSchaeferjohann: adding todo

zk.js/src/utxo/utxo.ts
```diff
@@ -14,77 +14,134 @@ import {
...
+/** Type safety: enforce that the utxo is not encrypted */
+export type Public = { isPublic: true };
+export type PublicBaseUtxo = BaseUtxo & Public;
+export type PublicUtxo = Utxo & Public;
+export type PublicOutUtxo = OutUtxo & Public;
+
+export type NullifierInputs = {
+  signature: BN;
+  /** hash of the utxo preimage */
+  hash: BN31;
```
ananas-block, line 109: I like the idea of this type. To make it better I believe we should make call it BN254 and extend constructor to check that the number is smaller than FIELD_SIZE eg.
```
BN254 extends BN
constructor() {
super()
let input = new BN(input)
if (input.gte(FIELD_SIZE)
  throw Error()
}
```

zk.js/src/utxo/utxo.ts
```diff
@@ -14,77 +14,134 @@ import {
...
-  encryptionPublicKey?: Uint8Array; // is only set if the utxo should be sent to another public key and thus be encrypted asymetrically
+import { BN31, createBN31 } from "./bn31";
+import {
+  PlaceHolderTData,
+  ProgramOutUtxo,
+  ProgramUtxo,
+  programOutUtxoToBytes,
+} from "./program-utxo";
+
+export const randomBN = (nbytes = 30) => new BN(nacl.randomBytes(nbytes));
```
ananas-block, line 32: this

zk.js/src/utxo/utxo.ts
```diff
@@ -96,107 +153,145 @@ export function createOutUtxo({
...
-    publicKey: utxoHashInputs.publicKey,
+}
+
+/** Utxo only */
+export function createOutUtxo({
+  owner,
+  amounts,
+  assets,
+  lightWasm,
+  blinding = new BN(randomBN(), 31, "be"),
```
ananas-block, line 213: randomBN() creates 30 bytes

zk.js/src/utxo/utxo.ts
```diff
@@ -96,107 +153,145 @@ export function createOutUtxo({
...
     poolType,
-    transactionVersion,
-    utxoDataHash,
+    version,
+    dataHash,
     metaHash,
     address,
   } = utxoHashInputs;
+
+  /// TODO: we're assuming that amount is < 2^254. Check security implications
```
ananas-block, line 281: we are checking that amounts are 64 bit numbers in circuit

zk.js/src/utxo/utxo.ts
```diff
@@ -672,113 +747,96 @@ export function createUtxo(
...
+    merkletreeId: BN_0.toNumber(), // TODO: make merkletreeId dynamic (support parallel merkle trees)
+    type,
     metaHash,
     address,
+    isPublic: false, // TODO: make isPublic dynamic
   };
   return utxo;
 }
 
+/// TODO: needs to be refactored to account for no-account access and public utxos
```
ananas-block, line 811: this will need to be exposed by the wallet since you need private key material to compute the signature(hash)

zk.js/tests/program-utxo.test.ts
```diff
@@ -45,22 +50,24 @@ describe("Program Utxo Functional", () => {
...
       assets: [SystemProgram.programId, MINT],
-      verifierAddress: pspId,
     };
   });
 
   it("create program out utxo", async () => {
     for (let i = 0; i < 100; i++) {
       const programOutUtxo = createProgramOutUtxo({
-        ...createOutUtxoInputs,
+        // owner: createOutUtxoInputs.owner,
```
ananas-block, line 62: leftover comment?

zk.js/src/build-balance.ts
SwenSchaeferjohann: good catch, will be redundant changes in follow up pr

zk.js/tests/program-utxo.test.ts
```diff
@@ -167,84 +168,91 @@ describe("Program Utxo Functional", () => {
...
       assets: [SystemProgram.programId, MINT],
-      verifierAddress: pspId,
+      verifierAddress: owner,
       blinding: new BN(new Uint8Array(31).fill(2)),
     };
     const assetLookupTable = lightProvider.lookUpTables.assetLookupTable;
 
     const programOutUtxo = createProgramOutUtxo({
-      ...createOutUtxoInputs,
+      // ...createOutUtxoInputs,
```
ananas-block, line 177: leftover comment?

zk.js/tests/program-utxo.test.ts
```diff
@@ -298,82 +305,102 @@ describe("Program Utxo Functional", () => {
...
+      lightWasm,
+    );
+
+    /// TODO: check whether this has downstream implications
     assert.equal(
-      decryptedUtxo.value?.utxo.nullifier,
-      "6375266098703930269345700550425174694499182299695686843103201658599361581059",
+      decryptedUtxo.value?.nullifier.toString(),
+      nullifier.toString(),
+      // "6375266098703930269345700550425174694499182299695686843103201658599361581059",
```
ananas-block, line 386: Please replace hardcoded value. The value is hardcoded to highlight that something changed which changed the value which usually should not be the case. Computing the comparison value dynamically defeats the purpose.

zk.js/tests/utxo.test.ts
```diff
@@ -162,17 +168,18 @@ describe("Utxo Functional", () => {
...
       hashAndTruncateToCircuit(SystemProgram.programId.toBytes()).toString(),
     );
     assert.equal(
-      outUtxo.assetsCircuit[1].toString(),
+      outUtxoAssetCircuitInput[1].toString(),
       hashAndTruncateToCircuit(assetPubkey.toBytes()).toString(),
     );
-    assert.equal(outUtxo.utxoDataHash.toString(), "0");
+    if ("data" in outUtxo) throw new Error("dataHash is not 0");
+    // assert.equal(outUtxo.dataHash.toString(), "0");
```
ananas-block, line 179: leftover comment?

zk.js/tests/utxo.test.ts
```diff
@@ -237,27 +244,30 @@ describe("Utxo Functional", () => {
...
+      decryptedUtxoAssetCircuitInput[0].toString(),
       hashAndTruncateToCircuit(SystemProgram.programId.toBytes()).toString(),
     );
     assert.equal(
-      decryptedUtxo.value?.assetsCircuit[1].toString(),
+      decryptedUtxoAssetCircuitInput[1].toString(),
       hashAndTruncateToCircuit(assetPubkey.toBytes()).toString(),
     );
-    assert.equal(decryptedUtxo.value?.utxoDataHash.toString(), "0");
+    // assert.equal(decryptedUtxo.value?.dataHash.toString(), "0");
```
ananas-block, line 258: leftover comment

zk.js/tests/verifiers.test.ts
```diff
@@ -246,42 +250,47 @@ describe("Verifier tests", () => {
...
+      owner: getVerifierProgramId(IDL_PUBLIC_LIGHT_PSP2IN2OUT),
       metaHash: new BN(6),
       address: new BN(7),
-      utxoDataHash: BN_1,
-      utxoData: { rnd: 1 },
+      data: { rnd: 1 },
+      ownerIdl: IDL_PUBLIC_LIGHT_PSP2IN2OUT,
+      /// Allows for custom hashing schema
+      dataHash: BN_1,
+      // dataHash: createDataHashWithDefaultHashingSchema({ rnd: 1 }, lightWasm),
```
ananas-block, line 285: where does it get the dataHash from?

ananas-block: shouldn't this throw an error?

SwenSchaeferjohann: we're passing BN_1 as dataHash and {rnd:1} as data; then BN_1 represents a custom hashing in this case

zk.js/src/utxo/select-in-utxos.ts
SwenSchaeferjohann: good point. this is quite intertwined with gettxparams/user class; will address when removing user class/taking care of the private txs side

ananas-block (approved):

PR #473, opened 2024-02-16
feat: merkle inclusion arkworks prover with groth16-solana outputs

sergeytimoshin:


circuit-lib/circuitlib-rs/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "circuitlib-rs"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+# light local deps
+# ark-zkey = { path = "../../../mopro/ark-zkey" }
+# witness = { path = "../../../circom-witness-rs" }
```
ananas-block, line 10: leftover comment

circuit-lib/circuitlib-rs/circuits/merkle/MerkleTreeProof.circom
```diff
@@ -0,0 +1,49 @@
+pragma circom 2.1.4;
+
+include "../../node_modules/circomlib/circuits/poseidon.circom";
+include "../../node_modules/circomlib/circuits/comparators.circom";
+include "../../node_modules/circomlib/circuits/bitify.circom";
+include "../../node_modules/circomlib/circuits/switcher.circom";
```
ananas-block, line 6: I think it would be cleaner to put the circuits into circuit-lib.circom, then we don't need to have the build script twice as well.
On the imports, you can use the -l option of the circom compiler to link folders to import from (see circuit-lib.circom).
`` circom example.circom -l node_modules/circom-lib/circuits/poseidon``

circuit-lib/circuitlib-rs/circuits/merkle/MerkleTreeProof.circom
```diff
@@ -0,0 +1,49 @@
...
+
+template MerkleTreeProof(levels, numberOfUTXOs) {
```
ananas-block, line 8: we have this template in circuit lib right?

ananas-block: in the Merkle tree folder I believe

circuit-lib/circuitlib-rs/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "circuitlib-rs",
```
ananas-block, line 2: if we move the circuits to circuit-lib.circom we probably don't need a package json in circuit-lib.rs

circuit-lib/circuitlib-rs/parse_vk_to_rust.js
```diff
@@ -0,0 +1,133 @@
+var ffjavascript = require('ffjavascript');
+const {unstringifyBigInts, leInt2Buff} = ffjavascript.utils;
+var fs = require("fs");
+const process = require('process');
```
ananas-block, line 4: we have this function in zk.js and an example of using it in circuit-lib.circom

circuit-lib/circuitlib-rs/src/groth16_solana_verifier.rs
```diff
@@ -0,0 +1,133 @@
...
+) -> Result<ProofResult, CircuitsError> {
+    let merkle_proof_inputs_len = merkle_proof_inputs.len();
+
+    let proof_inputs = ProofInputs(merkle_proof_inputs);
+    let inputs_hashmap: HashMap<String, Inputs> = proof_inputs
+        .try_into()
+        .map_err(|_| CircuitsError::ChangeEndiannessError)?;
+
+    let path = format!(
+        "test-data/merkle{}_{}/circuit.zkey",
```
ananas-block, line 61: can we pass in the path?
(I think when using this in tests the path is going to be different, so this won't work.)

circuit-lib/circuitlib-rs/src/merkle_proof_inputs.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl MerkleTreeInfo {
+    pub fn height(&self) -> u8 {
+        match self {
+            MerkleTreeInfo::H22 => 22,
+            MerkleTreeInfo::H30 => 30,
+        }
+    }
+    pub fn wasm_path(&self, num_of_utxos: usize) -> String {
```
ananas-block, line 19:
```suggestion
    pub fn test_wasm_path(&self, num_of_utxos: usize) -> String {
```

ananas-block (changes requested):

circuit-lib/circuitlib-rs/circuits/merkle/MerkleTreeProof.circom
sergeytimoshin: Yes, but it doesn't support variable number of utxos. Should we replace it new one?

circuit-lib/circuitlib-rs/parse_vk_to_rust.js
sergeytimoshin: zk.js is broken, so I moved the script to circom-lib.circom/scripts.

ananas-block (approved):

PR #493, opened 2024-03-04
chore: js with rust changes applied

SwenSchaeferjohann:
includes
rust:
- rpc_fee => relay_fee
- consistent CompressedProof struct 
(those are changes from @ananas-block that we merged into this feature branch directly)

js:
- adds stateless.js


js/stateless.js/examples/transfer-emit-events.ts
```diff
@@ -0,0 +1,142 @@
...
+import { confirmTx, sendAndConfirmTx } from "../src/test-utils";
+
+const FIXED_PAYER = byteArrayToKeypair([
+  122, 239, 192, 18, 21, 29, 237, 120, 104, 95, 247, 150, 181, 218, 207, 60,
+  158, 110, 200, 246, 74, 226, 30, 223, 142, 138, 133, 194, 30, 254, 132, 236,
+  227, 130, 162, 184, 215, 227, 81, 211, 134, 73, 118, 71, 219, 163, 243, 41,
+  118, 21, 155, 87, 11, 53, 153, 130, 178, 126, 151, 86, 225, 36, 251, 130,
+]);
+
+// <254bit ,
```
ananas-block, line 27:
```suggestion
// creates mock blinding < bn254 field size
```

js/stateless.js/src/constants.ts
```diff
@@ -0,0 +1,148 @@
...
+
+// TODO: make this for any
+// Change the constants that depend on LightSystemProgram.programId into functions
+
+export const getRegisteredProgramPda = () =>
+  new PublicKey("ytwwVWhQUMoTKdirKmvEW5xCRVr4B2dJZnToiHtE2L2"); // TODO: gov authority pda rn
+// PublicKey.findProgramAddressSync(
+//   [LightSystemProgram.programId.toBytes()],
+//   new PublicKey(accountCompressionProgram)
+// )[0];
```
ananas-block, line 34: should clean comments like this

js/stateless.js/src/state/utxo.ts
```diff
@@ -0,0 +1,228 @@
...
+   * TODO: Implement address functionality
+   * Optional unique account ID that is persistent across transactions.
+   */
+  address?: PublicKey;
+};
+
+/** Context for utxos inserted into a state Merkle tree */
+export type MerkleContext = {
+  /** Poseidon hash of the utxo preimage  */
+  hash: bigint254;
```
ananas-block, line 28: could consider it renaming to leaf since we have leafIndex below.

js/stateless.js/src/state/utxo.ts
```diff
@@ -0,0 +1,228 @@
...
+export type MerkleUpdateContext = {
+  /** Slot that the utxo was appended at */
+  slotUpdated: number;
+  /** Sequence */
+  seq: number;
+};
+
+export type MerkleContextWithMerkleProof = MerkleContext & {
+  /** Recent valid 'hash' proof path, expires after n slots */
+  merkleProof: PublicKey254[];
```
ananas-block, line 46: why PublicKey254 ?

ananas-block: I know they do that in compressed nfts but for zkps we need these to be big numbers not public keys

js/stateless.js/src/state/utxo.ts
SwenSchaeferjohann: adding into inline comment for now, will ask helius for thoughts since this will affect rpc interface

js/stateless.js/src/state/utxo.ts
SwenSchaeferjohann: replaced Publickey254 with BN254 (renamed from bigint254 bc there's no point in using bigint -which ATM is incompatible with anchor- as long as we're depending on BN.js indirectly. both anchor and web3js (for its PublicKey class) depend on BN.js).

i believe both are switching to native bigint in the midterm. we can reconsider then. 

we can also consider implementing a proper class for bn254 so that we can enforce the <254bit size at compile time (rn it's just a type that has the same shape as BN + runtime checkers).

SwenSchaeferjohann:
thx, addressed all comments

ananas-block (approved):

SwenSchaeferjohann:
thx for merging! restored branch for now so we don't nuke the links for investigating the failed CI runs.

PR #503, opened 2024-03-07
feat: add address to utxo

ananas-block:
Changes:
- added address option to utxo struct

Notes:
- we will want to switch to a type which is smaller than field size not Pubkey


sergeytimoshin (approved):

SwenSchaeferjohann (approved):
nice

PR #513, opened 2024-03-12
feat: JS compressed-token 

SwenSchaeferjohann:
- adds compressed-token JS client
- create_mint, mint_to
- adds JS CI workflows for compressed-token and stateless.js
- adds proper IDL generation and distribution (incl. fmt of idl .ts files)

note
- serde in client is not working yet, we'll merge that as a separate PR
- refactors separately


js/compressed-token/src/instructions/transfer.ts
```diff
@@ -0,0 +1,153 @@
...
+  is_native?: number;
+  /// The amount delegated
+  delegated_amount: number;
+  // TODO: validate that we don't need close authority
+  // /// Optional authority to close the account.
+  // close_authority?: PublicKey,
+};
+
+// FIXME:(swen) produces inconsistent remaining accounts with expected remaining accounts
+// this is currently akin to createExecuteCompressedInstruction
```
ananas-block, line 56: is this still the case?

js/compressed-token/src/program.ts
```diff
@@ -0,0 +1,241 @@
...
+   * @internal
+   */
+  constructor() {}
+
+  /**
+   * Public key that identifies the CompressedPda program
+   */
+  static programId: PublicKey = new PublicKey(
+    // TODO: can add check to ensure its consistent with the idl
+    '9sixVEthz2kMSKfeApZXHwuboT6DZuT6crAYJTciUCqE',
```
ananas-block, line 89: if we have it in constants as well probably makes sense to get if from there.

js/compressed-token/src/program.ts
```diff
@@ -0,0 +1,241 @@
...
+    return address;
+  }
+
+  static async createMint(
+    params: CreateMintParams,
+  ): Promise<TransactionInstruction[]> {
+    const mint = params.mint;
+    const authority = params.authority;
+    const feePayer = params.feePayer;
+    const rentExemptBalance = params.rentExemptBalance;
```
ananas-block, line 159: const {mint, authority, feePayer, rentExemptBalance } = params;

js/compressed-token/src/program.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  static async mintTo(params: MintToParams): Promise<TransactionInstruction> {
+    const systemKeys = defaultStaticAccountsStruct();
+
+    const { mint, feePayer, authority, merkleTree, toPubkey, amount } = params;
+
+    const tokenPoolPda = this.deriveTokenPoolPda(mint);
+    const mintAuthorityPda = this.deriveMintAuthorityPda(authority, mint);
+
+    const amounts = toArray(amount).map((amount) => bn(amount));
```
ananas-block, line 216: why do you need to do toArray here?

programs/compressed-token/src/process_transfer.rs
```diff
@@ -454,7 +479,9 @@ pub mod transfer_sdk {
             .iter()
             .map(|(k, _)| k.clone())
             .collect::<Vec<AccountMeta>>();
-
+        println!("_in_utxos {:?}", _in_utxos);
+        println!("out_utxos {:?}", out_utxos);
+        println!("remaining_accounts {:?}", remaining_accounts);
```
ananas-block, line 484: prints

programs/compressed-token/tests/test.rs
```diff
@@ -277,6 +277,9 @@ async fn test_transfer() {
     let in_utxos_tlv = mock_indexer.token_utxos[0].token_data.clone();
     let in_utxos = vec![mock_indexer.utxos[mock_indexer.token_utxos[0].index].clone()];
 
+    println!("in_utxos_tlv: {:?}", in_utxos_tlv);
+    println!("in_utxos: {:?}", in_utxos);
```
ananas-block, line 281: more prints

js/compressed-token/src/program.ts
SwenSchaeferjohann: for CompressedPdaProgram and CompressedTokenProgram we dont have it in constants since we have the program.ts classes, for these IMO best to just define it here _or_ get it directly from the IDL 

i.e. all consumers will just call `CompressedTokenProgram.programId` to get the value

js/compressed-token/src/program.ts
SwenSchaeferjohann: yup it's typesafe, it can either be one amount or an array of amounts in case you're batch-minting.
we're allowing non-arrays for  convenience when you mint a single instance

js/compressed-token/src/instructions/transfer.ts
SwenSchaeferjohann: nope, thx

ananas-block (approved):

PR #519, opened 2024-03-14
refactor: remove all cmds from cli except test-validator and config

sergeytimoshin:


cli/src/psp-utils/download.ts
```diff
@@ -5,8 +5,7 @@ import cliProgress from "cli-progress";
 import * as path from "path";
 import * as tar from "tar";
 import * as zlib from "zlib";
-import { sleep, getSystem, System } from "@lightprotocol/zk.js";
-import { CARGO_GENERATE_TAG } from "./constants";
+import { sleep } from "@lightprotocol/zk.js";
```
ananas-block, line 8: can we remove zk.js as dependency?

sergeytimoshin: Done.

sergeytimoshin: Unfortunately it's not as easy as it looks:
1. In the cli we use oclif-core: ^2, which is cjs, there is no esm support.
2. stateless.js is only esm, there is no backward compatibility with cjs.

So we have 2 options:
1. Migrate to oclif-core version 3, which has ESM support.
2. Add bundling to stateless.js.

Each of these options deserves to be on a different PR.  :)

ananas-block: open to either one which ever you think is faster

ananas-block (approved):

PR #524, opened 2024-03-15
feat: add create-mint, mint-to & transfer CLI commands.

sergeytimoshin:
New CLI commands create-mint, mint-to, and transfer have been added:
1. `create-mint [--mint-keypair <value>] [--mint-authority <value>]
    [--mint-decimals <value>]`: create a new compressed token mint.
2. `mint-to --mint <value> --to <value> --amount <value>
    [--mint-authority <value>]`:  mint tokens to an account.
3. `transfer --to <value> --amount <value> [--mint <value>]
    [--fee-payer <value>]`: transfer tokens from one account to another.


cli/src/commands/create-mint/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    loader.start();
+
+    try {
+      let mintDecimals = flags["mintDecimals"];
+      if (!mintDecimals) {
+        mintDecimals = 2;
+      }
+      const payer = getPayer();
+
+      const randomMint = Keypair.generate();
```
SwenSchaeferjohann, line 48: for users who don't want to pass in a specific mint keypair into cmd args, `createMint() `already creates a random mint keypair on the fly under the hood, so you can just pass `undefined`

cli/src/commands/mint-to/index.ts
```diff
@@ -0,0 +1,80 @@
...
+      throw new Error("Invalid arguments");
+    }
+
+    const loader = new CustomLoader(`Performing mint-to...\n`);
+    loader.start();
+
+    try {
+      const mintPublicKey = mint
+        ? new PublicKey(mint)
+        : Keypair.generate().publicKey;
```
SwenSchaeferjohann, line 54: note that this should be the mintAuthority - so users will always have to pass this into the cmd. 
(i think by default for the equivalent cmd in the solana cli, they read from the default solana keypair (i.e. by default they assume that `payer` and `mintAuthority` are the same (which they usually are but that depends on what you pass as authority into `createMint`

cli/src/commands/create-mint/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      const connection = new Connection(getSolanaRpcUrl());
+
+      // TODO: remove this, it's just for testing
+      const sig = await connection.requestAirdrop(payer.publicKey, 2e9);
+      await confirmTx(connection, sig);
+
+      const { mint, transactionSignature } = await createMint(
+        connection,
+        payer,
+        payer.publicKey,
```
SwenSchaeferjohann, line 58: we should allow users to pass in a separate `mintAuthority` (by default this would usually be the same as `payer`, which in turn should be read from the default destination keypair.json like solana cli does it iirc)

cli/src/commands/mint-to/index.ts
```diff
@@ -0,0 +1,80 @@
...
+      // TODO: remove this, it's just for testing
+      const sig = await connection.requestAirdrop(payer.publicKey, 2e9);
+      await confirmTx(connection, sig);
+
+      const txId = await mintTo(
+        connection,
+        payer,
+        mintPublicKey,
+        toPublicKey,
+        payer.publicKey,
```
SwenSchaeferjohann, line 68: note that this should be the mintAuthority - so users will always have to pass this into the cmd.
(i think by default for the equivalent cmd in the solana cli, they read from the default solana keypair (i.e. by default they assume that payer and mintAuthority are the same (which they usually are but that depends on what you pass as authority into createMint

cli/src/commands/mint-to/index.ts
SwenSchaeferjohann, line 54: would it make sense to always require to pass the mint? i think generating a new mintpubkey just wont work bc the specific mint needs to have been created first

cli/src/commands/transfer/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      // TODO: remove this, it's just for testing
+      const sig = await connection.requestAirdrop(payer.publicKey, 2e9);
+      await confirmTx(connection, sig);
+
+      const txId = await transfer(
+        connection,
+        payer,
+        mintPublicKey,
+        amount,
+        payer,
```
SwenSchaeferjohann, line 67: think we should allow users to pass a separate feepayer than owner

SwenSchaeferjohann (commented):
left few thoughts on things that i'm sure you were already aware of 👍

vadorovsky (approved):
LGTM

SwenSchaeferjohann (approved):
- not sure mintTo will work if mintAuthority != payer. need to pass the keypair into mintTo in that case @sergeytimoshin 
- let's add a test for this in a follow up pr

PR #526, opened 2024-03-16
feat: spend real input utxos in transfer

SwenSchaeferjohann:


js/compressed-token/src/actions/transfer.ts
```diff
@@ -6,30 +6,56 @@ import {
...
+ * Transfer compressed tokens from one owner to another
  *
  * @param connection     Connection to use
  * @param payer          Payer of the transaction fees
  * @param mint           Mint of the compressed token
  * @param amount         Number of tokens to transfer
- * @param currentOwner   Source address
- * @param newOwner       Destination address
+ * @param owner          Owner of the compressed tokens
+ * @param toAddress      Destination address of the recipient
```
ananas-block, line 58: I think address is confusing because we use address as pda address.
What about toOwner or toPubkey?

js/compressed-token/src/token-serde.ts
```diff
@@ -0,0 +1,150 @@
...
+export function parseTokenLayoutWithIdl(
+  utxo: Utxo_IdlType,
+): TokenTlvData_IdlType | null {
+  if (utxo.data === null) {
+    return null;
+  }
+
+  if (utxo.data.tlvElements.length === 0) return null;
+  if (
+    /// We can assume 0th element is token
```
ananas-block, line 38: Is this a good assumption?
What if we have custom pdas with different data, then the process crashes right?

ananas-block (approved):

PR #540, opened 2024-03-18
fix: use $HOME/.config/solana/id.json keypair as a default payer in CLI

sergeytimoshin:


cli/test/commands/transfer/index.test.ts
```diff
@@ -1,19 +1,24 @@
...
-
+    const mintKeypair = defaultSolanaWalletKeypair() || Keypair.generate();
+    await requestAirdrop(mintKeypair.publicKey);
     const mintAmount = 10;
-    const mintAuthority = getPayer().publicKey;
-    const mintDestination = getPayer().publicKey;
-    const mintAddress = await createTestMint();
+    const mintAuthority = mintKeypair.publicKey;
+    const mintDestination = mintKeypair.publicKey;
+    const mintAddress = await createTestMint(mintKeypair);
```
ananas-block, line 19: I think the mintAuthority and mintKeypair should be different keypairs so that it is realistic.
(The mint keypair will become a Solana account. You don't want to turn your wallet keypair into a mint account.)

cli/test/commands/transfer/index.test.ts
```diff
@@ -1,19 +1,24 @@
...
+} from "../../../src";
+import { Connection, Keypair, PublicKey } from "@solana/web3.js";
 import { createMint, mintTo } from "@lightprotocol/compressed-token";
 import { requestAirdrop } from "../../helpers/helpers";
 describe("transfer", () => {
   test.it(async () => {
     await initTestEnvIfNeeded();
-    await requestAirdrop(getPayer().publicKey);
-
+    const mintKeypair = defaultSolanaWalletKeypair() || Keypair.generate();
```
ananas-block, line 14:
```suggestion
    const mintKeypair = Keypair.generate();
    const payerKeypair = defaultSolanaWalletKeypair();
```

cli/test/commands/create-mint/index.test.ts
```diff
@@ -1,17 +1,16 @@
...
     await initTestEnvIfNeeded();
-    await requestAirdrop(getPayer().publicKey);
-
-    const mintAuthority = getPayer().publicKey.toBase58();
     const mintDecimals = 5;
-    const mintKeypair = getPayer();
-
+    const mintKeypair = defaultSolanaWalletKeypair() || Keypair.generate();
+    const mintAuthority = mintKeypair.publicKey.toBase58();
+    await requestAirdrop(mintKeypair.publicKey);
```
ananas-block, line 13: I think the mintAuthority and mintKeypair should be different keypairs so that it is realistic.
(The mint keypair will become a Solana account. You don't want to turn your wallet keypair into a mint account.)

cli/test/commands/create-mint/index.test.ts
```diff
@@ -1,24 +1,24 @@
...
 describe("create-mint", () => {
   test.it(async () => {
     await initTestEnvIfNeeded();
-    await requestAirdrop(getPayer().publicKey);
-
-    const mintAuthority = getPayer().publicKey.toBase58();
     const mintDecimals = 5;
-    const mintKeypair = getPayer();
-
+    const mintKeypair = defaultSolanaWalletKeypair();
```
SwenSchaeferjohann, line 11: using the solana wallet as mintkeypair is unintended (this would cause the user's pubkey to become a Mint address), but i think its non critical. we can add more test cases in followup prs imo.

cli/test/commands/mint-to/index.test.ts
```diff
@@ -1,18 +1,22 @@
...
 describe("mint-to", () => {
   test.it(async () => {
     await initTestEnvIfNeeded();
-    await requestAirdrop(getPayer().publicKey);
-
+    const mintKeypair = defaultSolanaWalletKeypair() || Keypair.generate();
+    await requestAirdrop(mintKeypair.publicKey);
     const mintAmount = 100;
-    const mintAuthority = getPayer().publicKey.toBase58();
+    const mintAuthority = mintKeypair.publicKey.toBase58();
```
SwenSchaeferjohann, line 19: same goes for this -  using the mintKeypair as authority is not intended but it works in this case because create_mint sets the mintkeypair as authority. we can add other testcases in follow up PRs

cli/test/commands/transfer/index.test.ts
```diff
@@ -1,47 +1,54 @@
...
-        `transfer ${mintAmount} tokens to ${mintDestination.toBase58()} from ${mintAddress.toBase58()}, fee-payer: ${mintAuthority.toBase58()} `,
+        `transfer ${mintAmount} tokens to ${mintDestination.toBase58()} from ${mintAddress.toBase58()}, fee-payer: ${payerKeypair.publicKey.toBase58()} `,
         (ctx: any) => {
           expect(ctx.stdout).to.contain("mint-to successful");
         },
       );
   });
 
-  async function createTestMint() {
+  async function createTestMint(payer: Keypair) {
```
SwenSchaeferjohann, line 46: same for this - non critical, so its gtg. let's add more test cases for payer != authority (and neither being equal to the mintkeypair)

SwenSchaeferjohann (approved):
gtg because technically the tests are fine, we'll have to add more unit test coverage (also including failure cases) in followup PRs anyway

PR #541, opened 2024-03-19
test: add gnark inclusion unit tests

sergeytimoshin:
Included test data for an inclusion proof in a new file "inclusion.csv". 
Tests are automated and fail/ pass status is checked based on the expected results from the csv test data.


SwenSchaeferjohann (approved):
great thanks!

PR #543, opened 2024-03-20
feat: cli: get keypairs by file paths when using create-mint, mint-to & transfer commands

sergeytimoshin:
Flags updated:
1. create-mint: --mint-keypair flag expects file path to the keypair (optional).
2. mint-to: --mint-authority flag expects file path to the keypair (optional, defaults to $HOME/.config/solana/id.json).
3. transfer: --fee-payer flag expects file path to the keypair (optional, defaults to $HOME/.config/solana/id.json).


SwenSchaeferjohann (approved):
Great thx!

PR #544, opened 2024-03-20
refactor: utxo to compressed account

ananas-block:
Changes:
- replaced utxo with compressed account both in variables and comments
- refactored tlv
- commented serialized utxos and optimized instruction


js/stateless.js/src/idls/psp_compressed_pda.ts
```diff
@@ -415,272 +218,94 @@ export type PspCompressedPda = {
...
             type: {
               vec: {
-                defined: 'InUtxoSerializableTuple';
+                array: ['u8', 32];
               };
             };
           },
           {
-            name: 'outUtxos';
+            name: 'outputAccountHashes';
```
SwenSchaeferjohann, line 234: inconsistency w input equivalent, 

`outputAccountHashes` -> `outputCompressedAccountHashes`

SwenSchaeferjohann (approved):
found some nits. not urgent but sharing for visibility regardless

js/stateless.js/src/idls/psp_compressed_pda.ts
```diff
@@ -145,249 +98,103 @@ export type PspCompressedPda = {
...
-            type: {
-              option: 'u64';
-            };
-          },
-          {
-            name: 'outUtxoIndices';
-            type: {
-              vec: 'u64';
-            };
+            name: 'indexMtAccount';
```
SwenSchaeferjohann, line 152: elsewhere a similar value is used in `outputStateMerkleTreeAccountIndices` fwiw

programs/compressed-pda/src/instructions.rs
```diff
@@ -124,60 +155,56 @@ pub struct TransferInstruction<'info> {
...
-#[derive(Debug, Clone, AnchorSerialize, AnchorDeserialize)]
-pub struct CpiSignature {
-    pub program: Pubkey,
-    pub tlv_hash: [u8; 32],
-    pub tlv_data: TlvDataElement,
-}
-
-// TODO: parse utxos a more efficient way, since owner is sent multiple times this way
 #[derive(Debug)]
 #[account]
```
SwenSchaeferjohann, line 167: the struct InstructionDataTransfer is now used by the `cpiSignatureAccount`, so tagging it as anchor account breaks the IDL.
instead, we can tag it as regular type with AnchorDeserialize,also allows us to not have to prepend 8 bytes when deserializing and can use ::deserialize instead of try_deserialize_unchecked()

programs/compressed-token/src/process_transfer.rs
```diff
@@ -243,19 +241,20 @@ pub struct TransferInstruction<'info> {
     pub self_program: Program<'info, crate::program::PspCompressedToken>,
 }
 
-// TODO: parse utxos a more efficient way, since owner is sent multiple times this way
+// TODO: parse compressed_accounts a more efficient way, since owner is sent multiple times this way
 // This struct is equivalent to the InstructionDataTransfer, but uses the imported types from the psp_compressed_pda
 #[derive(Debug, Clone, AnchorSerialize, AnchorDeserialize)]
 pub struct InstructionDataTransfer {
```
SwenSchaeferjohann, line 247: if we give this struct the same name as in compress-pda, we break the IDL. 
(previously it just didn't break because compressed-pda was defining its InstructionDataTransfer struct as account not as type)
In my feature branch,  I've temporarily renamed this to CompressedTokenInstructionDataTransfer in order to fix the IDL. there's probably a better name for it tho.

programs/compressed-pda/src/compressed_account.rs
```diff
@@ -0,0 +1,201 @@
+use anchor_lang::prelude::*;
+use light_hasher::{Hasher, Poseidon};
+use light_utils::hash_to_bn254_field_size_le;
+
+#[derive(Debug, PartialEq, Default, Clone, AnchorSerialize, AnchorDeserialize)]
+pub struct CompressedAccountWithMerkleContext {
+    pub compressed_account: CompressedAccount,
+    pub index_mt_account: u8,
+    pub index_nullifier_array_account: u8,
+    pub leaf_index: u32,
```
SwenSchaeferjohann, line 10: turns out it was indeed using u32 all along onchain

SwenSchaeferjohann (commented):
left some more comments on what i found while working on #548  re: fixing IDL

programs/compressed-token/src/process_transfer.rs
```diff
@@ -243,19 +252,20 @@ pub struct TransferInstruction<'info> {
...
-    out_utxos: Vec<TokenTransferOutUtxo>,
+    input_compressed_accounts_with_merkle_context: Vec<CompressedAccountWithMerkleContext>,
+    input_token_data: Vec<TokenData>,
+    output_compressed_accounts: Vec<TokenTransferOutputData>,
+    output_state_merkle_tree_account_indices: Vec<u8>,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, AnchorSerialize, AnchorDeserialize)]
-pub struct TokenTransferOutUtxo {
+pub struct TokenTransferOutputData {
```
SwenSchaeferjohann, line 268: `TokenTransferOutputData` still has `index_mt_account` so we're currently sending the same data to the chain twice. once as part of the `TokenTransferOutputData`  and once in the  `output_state_merkle_tree_account_indices` of the InstructionDataTransfer of compressed-token in line 487

programs/compressed-pda/src/event.rs
```diff
@@ -5,15 +5,23 @@ use anchor_lang::{
...
-    pub in_utxos: Vec<Utxo>,
-    pub out_utxos: Vec<Utxo>,
-    pub de_compress_amount: Option<u64>,
-    pub out_utxo_indices: Vec<u64>,
+    pub input_compressed_account_hashes: Vec<[u8; 32]>,
+    pub output_account_hashes: Vec<[u8; 32]>,
+    pub input_compressed_accounts: Vec<CompressedAccountWithMerkleContext>,
+    pub output_compressed_accounts: Vec<CompressedAccount>,
+    // index of Merkle tree account in remaining accounts
+    pub output_state_merkle_tree_account_indices: Vec<u8>,
```
SwenSchaeferjohann, line 20: for these to be fully useful in the mock-rpc (i.e. client/mock-rpc doesn't have to fetch additional data) we'd also have to emit the `output_nullifier_queue_indices`
and put them into the pubkey_array... but onchain doesn't have access to that data AFAIK since its only needed for spending accounts. instead of finding a workaround for this we should prolly make the PublicTransactionEvent smaller again, not emit the output_queues, and not emit any of the redundant info at all either and just implement a few additional steps in the test-rpc.  cc @ananas-block

programs/compressed-token/src/process_transfer.rs
ananas-block: fixed

programs/compressed-pda/src/compressed_account.rs
```diff
@@ -0,0 +1,209 @@
+use anchor_lang::prelude::*;
+use light_hasher::{Hasher, Poseidon};
+use light_utils::hash_to_bn254_field_size_le;
+
+#[derive(Debug, PartialEq, Default, Clone, AnchorSerialize, AnchorDeserialize)]
+pub struct CompressedAccountWithMerkleContext {
+    pub compressed_account: CompressedAccount,
+    pub index_merkle_tree_account: u8,
+    pub index_nullifier_array_account: u8,
```
SwenSchaeferjohann, line 9: @ananas-block  should we just clean up the api now? i.e. `merkle_tree_account_index` and `nullifier_array_account_index` 
such that the naming is consistent with _index suffix in leaf_index. 

(and more consistent with use of nullifier queue elsewhere would be to do
`nullifier_queue_index`)

programs/compressed-pda/src/event.rs
```diff
@@ -5,15 +5,23 @@ use anchor_lang::{
...
-    pub out_utxo_indices: Vec<u64>,
+    pub input_compressed_account_hashes: Vec<[u8; 32]>,
+    pub output_compressed_account_hashes: Vec<[u8; 32]>,
+    pub input_compressed_accounts: Vec<CompressedAccountWithMerkleContext>,
+    pub output_compressed_accounts: Vec<CompressedAccount>,
+    // index of Merkle tree account in remaining accounts
+    pub output_state_merkle_tree_account_indices: Vec<u8>,
+    pub output_leaf_indices: Vec<u32>,
     pub relay_fee: Option<u64>,
+    pub de_compress_amount: Option<u64>,
```
SwenSchaeferjohann, line 23: @ananas-block  do we want to add the `is_compress`: bool now?

programs/compressed-pda/src/compressed_account.rs
ananas-block: sure

PR #547, opened 2024-03-22
feat: enable acp group access control

ananas-block:
Changes:
- account compression program:
  - enabled group access control
  - moved group tests to separate file
  - added failing tests for merkle tree append, nullify, insert into indexed array
  - linked merkle tree account and indexed array by pubkey and checking these in nullify instruction

### DIFF
--- a/programs/account-compression/src/instructions/insert_into_indexed_array.rs
+++ b/programs/account-compression/src/instructions/insert_into_indexed_array.rs
@@ -80,7 +80,7 @@ pub fn process_initialize_indexed_array<'info>(
     indexed_array_account.owner = owner;
     indexed_array_account.delegate = delegate.unwrap_or(owner);
     indexed_array_account.associated_merkle_tree = associated_merkle_tree.unwrap_or_default();
-    // Explicitly initializing the indexed array is not necessary as defautl values are all zero.
+    // Explicitly initializing the indexed array is not necessary as default values are all zero.
     Ok(())
 }
 
--- a/programs/account-compression/src/instructions/insert_into_indexed_array.rs
+++ b/programs/account-compression/src/instructions/insert_into_indexed_array.rs
@@ -4,7 +4,13 @@ use aligned_sized::aligned_sized;
 use anchor_lang::{prelude::*, solana_program::pubkey::Pubkey};
 use bytemuck::{Pod, Zeroable};
 
-use crate::{utils::constants::STATE_INDEXED_ARRAY_SIZE, RegisteredProgram};
+use crate::{
+    utils::{
+        check_registered_or_signer::{check_registered_or_signer, GroupAccess, GroupAccounts},
+        constants::STATE_INDEXED_ARRAY_SIZE,
+    },
+    RegisteredProgram,
+};
 
 #[derive(Accounts)]
 pub struct InsertIntoIndexedArrays<'info> {
@@ -38,6 +44,7 @@ pub fn process_insert_into_indexed_arrays<'a, 'b, 'c: 'info, 'info>(
                 array_map.insert(mt.key(), (mt, Vec::new()));
             }
         };
+
         array_map.get_mut(&mt.key()).unwrap().1.push(elements[i]);
     }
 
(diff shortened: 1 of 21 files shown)


programs/account-compression/src/instructions/insert_into_indexed_array.rs
```diff
@@ -62,11 +73,13 @@ pub fn process_initialize_indexed_array<'info>(
...
     owner: Pubkey,
     delegate: Option<Pubkey>,
+    associated_merkle_tree: Option<Pubkey>,
 ) -> Result<()> {
     let mut indexed_array_account = ctx.accounts.indexed_array.load_init()?;
     indexed_array_account.index = index;
     indexed_array_account.owner = owner;
     indexed_array_account.delegate = delegate.unwrap_or(owner);
+    indexed_array_account.associated_merkle_tree = associated_merkle_tree.unwrap_or_default();
     // Explicitly initializing the indexed array is not necessary as defautl values are all zero.
```
SwenSchaeferjohann, line 83:
```suggestion
    // Explicitly initializing the indexed array is not necessary as default values are all zero.
```

programs/account-compression/src/instructions/nullify_leaves.rs
```diff
@@ -28,33 +28,63 @@ pub struct NullifyLeaves<'info> {
...
-    // check_registered_or_signer::<AppendLeaves, StateMerkleTreeAccount>(
-    //     &ctx,
-    //     &merkle_tree_account,
-    // )?;
     let mut array_account = ctx.accounts.indexed_array.load_mut()?;
-    let leaf: [u8; 32] = array_account.indexed_array[leaves_indices[0] as usize].element;
+    let mut merkle_tree_account = ctx.accounts.merkle_tree.load_mut()?;
+    if array_account.associated_merkle_tree != ctx.accounts.merkle_tree.key() {
+        msg!(
+            "merkle tree and indexed array are not associated {} != {}",
```
SwenSchaeferjohann, line 39: think this would rather be along the lines of 'provided merkle tree doesn't match associated merkle tree' ?

programs/account-compression/src/instructions/nullify_leaves.rs
```diff
@@ -28,33 +28,63 @@ pub struct NullifyLeaves<'info> {
...
+            array_account.associated_merkle_tree,
+            ctx.accounts.merkle_tree.key()
+        );
+        return Err(AccountCompressionErrorCode::InvalidMerkleTree.into());
+    }
+    if merkle_tree_account.associated_queue != ctx.accounts.indexed_array.key() {
+        msg!(
+            "merkle tree and indexed array are not associated {} != {}",
```
SwenSchaeferjohann, line 47: "provided merkle tree [...]" ?; also may make sense to consolidate the names for queue and indexed_array / or add a note saying that it's the same

programs/account-compression/tests/group_authority_tests.rs
```diff
@@ -0,0 +1,188 @@
...
+        &[instruction],
+        Some(&context.payer.pubkey()),
+        &vec![&context.payer],
+        context.last_blockhash,
+    );
+    context
+        .banks_client
+        .process_transaction(transaction)
+        .await
+        .unwrap_err();
```
SwenSchaeferjohann, line 187: i suspect `.unwrap_err()` implicitly expects this to return an error?

programs/compressed-pda/src/compressed_account.rs
```diff
@@ -52,6 +52,11 @@ impl CompressedAccount {
         }
 
         if let Some(data) = &self.data {
+            // TODO: double check that it is impossible to create a hash collisions for different sized poseidon hash inputs
+            // Otherwise we could use padding so prevent a theoretical attack producing a hash collision
```
SwenSchaeferjohann, line 56:
```suggestion
            // Otherwise we could use padding to prevent a theoretical attack producing a hash collision
```

SwenSchaeferjohann (approved):
lgmt, left some questions / typos above,

programs/account-compression/tests/group_authority_tests.rs
ananas-block: yes it expects to unwrap an error

programs/account-compression/src/instructions/nullify_leaves.rs
ananas-block: Actually I think since this transaction wants to nullify a value in a Merkle tree, the queue wouldn't be associated with the Merkle tree.

ananas-block:
Also just noticed that we need to check that every Merkle tree and queue can only be associated to one other tree or queue.

programs/account-compression/src/instructions/nullify_leaves.rs
ananas-block: Changed it in the logs, I want to do a separate pr to change the variables.

programs/account-compression/src/instructions/nullify_leaves.rs
ananas-block: ah now I got it adapted it.

PR #566, opened 2024-03-29
feat: interface with photon

SwenSchaeferjohann:
- needed for sdk<>photon and cli<>photon testing
- synced with spec
- fixes cli commands and cli tests, but doesn't include them in ci yet  (running `pnpm test` fails because the prover gets started the wrong way)


SwenSchaeferjohann:
fyi temporarily disabled `&& pnpm test:browser in prover.js` i think version mismatches caused an issue there.

PR #567, opened 2024-03-29
feat: sol decompression

ananas-block:
Changes:
- add sol compression and decompression


SwenSchaeferjohann (commented):
lgtm. should we find an alternative for `de_compress` right away?
suggestions:
`compress_decompress_lamports` and
`fn compress_decompress_lamports()`
and so forth.

yet another alternative may be 
`compress_or_decompress_lamports`

SwenSchaeferjohann:
shouldn't CI for the JS tests fail here since other nullable values ought to be encoded and decoded on the client? or does anchor auto-fill for appended nullable properties? @ananas-block

test-utils/src/test_env.rs
```diff
@@ -228,6 +228,25 @@ pub async fn setup_test_programs_with_accounts() -> EnvWithAccounts {
...
+    //     };
+    //     let instruction = Instruction {
+    //         program_id: psp_compressed_pda::ID,
+    //         accounts: accounts.to_account_metas(Some(true)),
+    //         data: instruction_data.data(),
+    //     };
+    //     create_and_send_transaction(&mut context, &[instruction], &payer.pubkey(), &[&payer])
+    //         .await
+    //         .unwrap();
+    // }
```
SwenSchaeferjohann, line 249: can we add a comment/todo above here?

SwenSchaeferjohann (approved):

PR #568, opened 2024-03-29
feat: decompress spl

ananas-block:
Changes:
- rename ``de_compress``  to ``compression``
- token cpi authority pda derivation seeds from `` [b"cpi_authority"]`` to ``[[b"cpi_authority", account_compression::ID.to_bytes().as_slice()]``
- add spl compression and decompression


SwenSchaeferjohann (commented):
can we change compression_lamports to compress_or_decompress_lamports and so on? imo explicit and descriptive is best here even at the expense of word length

unless otherwise discussed i will then include that name change in my compress-spl pr which we'll merge into this one here. wdyt? @ananas-block

PR #573, opened 2024-04-02
ci: fix: js.yml. Remove 'with' command under 'Checkout sources'.

sergeytimoshin:


SwenSchaeferjohann:
thanks. closing as it is fixed in the most recently merged pr already

PR #580, opened 2024-04-06
feat: add SOL compression and decompression cli commands

sergeytimoshin:
New CLI commands `compress-sol` and `decompress-sol` have been added:

• `compress-sol --amount <value> --to <value>`: compress SOL to an account.
• `decompress-sol --amount <value> --to <value>`: decompress SOL to an account.

### DIFF
--- a/cli/src/commands/decompress-sol/index.ts
+++ b/cli/src/commands/decompress-sol/index.ts
@@ -6,7 +6,11 @@ import {
   getSolanaRpcUrl,
 } from "../../utils/utils";
 import { Connection, PublicKey, LAMPORTS_PER_SOL } from "@solana/web3.js";
-import { Rpc, decompressLamports } from "@lightprotocol/stateless.js";
+import {
+  Rpc,
+  decompressLamports,
+  getTestRpc,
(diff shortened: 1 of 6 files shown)


cli/src/commands/decompress-sol/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    const loader = new CustomLoader(`Performing compress-sol...\n`);
+    loader.start();
+
+    try {
+      const toPublicKey = new PublicKey(to);
+      const payer = defaultSolanaWalletKeypair();
+
+      const connection = new Connection(getSolanaRpcUrl());
```
SwenSchaeferjohann, line 44: needs to be `getTestRpc` / new `TestRpc`() instance, which extends `Connection` with the compression API methods.
(we'll then independently have to replace `TestRpc` use with `Rpc`  once we've successfully tested Rpc class with local photon+sdk)

SwenSchaeferjohann (changes requested):

SwenSchaeferjohann (approved):
thx!

PR #587, opened 2024-04-11
feat: run the photon indexer using the --photon flag (default=true) on the start-local-validator cli cmd.

sergeytimoshin:


sergeytimoshin:
Blocked by https://github.com/Lightprotocol/light-protocol/pull/588

cli/src/commands/test-validator/index.ts
```diff
@@ -43,7 +30,7 @@ class SetupCommand extends Command {
     loader.start();
     await initTestEnv({
       skipSystemAccounts: flags.skip_system_accounts,
-      background: flags.background,
```
SwenSchaeferjohann, line 46: why rm background flag?

sergeytimoshin: We always use background, it's enabled by default now. Do we really need blocking foreground process?

sergeytimoshin: One more reason: start-local-validator can spawn several different processes (validator, photon, etc) at the same time, so this background flag can be confusing, because all of them except validator would be in background anyway.

SwenSchaeferjohann: yeah makes sense. previously though we actually ran it in foreground by default 
(background: Flags.boolean({
      char: "b",
      description: "Runs a test validator as a process in the background.",
      default: false,)
      
     just wondering whether the switch to default background causes the issue?

SwenSchaeferjohann: i.e. the default command would override the default value of initTestEnv every time, unless the background flag was explicitly passed with background=true

SwenSchaeferjohann:
lfg.

SwenSchaeferjohann (approved):

PR #591, opened 2024-04-11
feat: add bool flags for enabling prover circuits

sergeytimoshin:
```
FLAGS
  -i, --without-indexer        Runs a test validator without indexer service.
  -p, --without-prover         Runs a test validator without prover service.
  -s, --skip-system-accounts   Runs a test validator without initialized light system accounts.
  --prove-compressed-accounts  Enable proving of compressed accounts.
  --prove-new-addresses        Enable proving of new addresses.

```


SwenSchaeferjohann (approved):

SwenSchaeferjohann (commented):
lgtm once ci green

cli/src/commands/test-validator/index.ts
```diff
@@ -10,22 +10,32 @@ class SetupCommand extends Command {
...
       default: false,
     }),
+    "prove-compressed-accounts": Flags.boolean({
+      description: "Enable proving of compressed accounts.",
+      default: true,
+      exclusive: ["without-prover"],
+    }),
+    "prove-new-addresses": Flags.boolean({
+      description: "Enable proving of new addresses.",
+      default: false,
```
ananas-block, line 36: I think this should be true by default as well, otherwise many people are going to be stuck immediately.

PR #592, opened 2024-04-12
fix: cli tests

sergeytimoshin:


cli/src/commands/compress-sol/index.ts
```diff
@@ -18,7 +18,7 @@ class MintToCommand extends Command {
       description: "Specify the recipient address.",
       required: true,
     }),
-    amount: Flags.integer({
+    amount: Flags.string({
```
SwenSchaeferjohann, line 21: why string?

cli/test/commands/create-mint/index.test.ts
```diff
@@ -5,23 +5,25 @@ import { requestAirdrop } from "../../helpers/helpers";
...
-        expect(ctx.stdout).to.contain("create-mint successful");
-      });
   });
+
+  test
+    .stdout()
+    .command([
+      "create-mint",
+      `--mint-decimals=${mintDecimals}`,
+      `--mint-authority=${mintAuthority.publicKey.toBase58()}`,
```
SwenSchaeferjohann, line 23: this should fail; create-mint requires mintAuthority to be a signer to work

cli/test/commands/mint-to/index.test.ts
```diff
@@ -5,52 +5,37 @@ import {
...
-      await getPayer(),
-      await getPayer(),
-      9,
+  test
+    .stdout()
+    .command([
+      "mint-to",
+      `--amount=${mintAmount}`,
+      `--mint=${mintAddress}`,
+      `--mint-authority=${mintAuthority}`,
```
SwenSchaeferjohann, line 32: mint-to action requires the authority to be a signer to work (since were not using multisigsigners)

PR #600, opened 2024-04-13
feat: we only need /prove endpoint

sergeytimoshin:
* remove /inclusion/, /noninclusion, /combined endpoints
* add /prove endpoint
* refactor input data format https://gist.github.com/sergeytimoshin/beec2246537a30bd5adb1b87b51ba893


gnark-prover/prover/marshal_non_inclusion.go
```diff
@@ -6,15 +6,18 @@ import (
...
-	InPathElements [][]string `json:"inPathElements"`
+type NonInclusionProofInputsJSON struct {
+	Root                 string   `json:"root"`
+	Value                string   `json:"value"`
+	PathIndex            uint32   `json:"pathIndex"`
+	PathElements         []string `json:"pathElements"`
+	LeafLowerRangeValue  string   `json:"leafLowerRangeValue"`
+	LeafHigherRangeValue string   `json:"leafHigherRangeValue"`
+	LeafIndex            uint32   `json:"leafIndex"`
+}
```
ananas-block, line 17: shouldn't there only be one ProofInputsJSON marshall type now sine we only have one endpoint?

gnark-prover/prover/combined_proving_system.go
```diff
@@ -16,23 +16,23 @@ type CombinedParameters struct {
 	NonInclusionParameters NonInclusionParameters
 }
 
-func (p *CombinedParameters) NumberOfUTXOs() uint32 {
+func (p *CombinedParameters) NumberOfUTXOs() int {
```
ananas-block, line 19: why switch to int?

ananas-block: (in case that it is unrelated to the single endpoint it would make reviewing a lot easier to do the switch in a different pr)

gnark-prover/server/server.go
```diff
@@ -122,39 +111,31 @@ func (handler inclusionHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	var proof *prover.Proof
+	var circuitType prover.CircuitType
 
-	var params prover.InclusionParameters
-
-	err = json.Unmarshal(buf, &params)
+	circuitType, err = prover.ParseCircuitType(buf)
```
ananas-block, line 116: why don't we just use t[he specified input Json ](https://gist.github.com/sergeytimoshin/beec2246537a30bd5adb1b87b51ba893)here and decide why circuit to use based on that?
```

if new addresses and input compressed accounts -> combined
else if input compressed accounts -> inclusion
else if new addresses -> non inclusion
```

gnark-prover/prover/marshal_non_inclusion.go
sergeytimoshin: Not necessarily, we compose combined circuit using decoupled inclusion/non-inclusion, I think this makes code easier to understand, less burden:

 https://github.com/Lightprotocol/light-protocol/pull/600/files#diff-8d6043aeb42b679af66348c1b85c2fbe129d39fba985e5859323b443e6f62ea2R9-R10

gnark-prover/server/server.go
sergeytimoshin: It's basically what we are doing there. :)

What I don't really like is that I still have ["inclusion" & "non-inclusion" flags for cli](https://github.com/Lightprotocol/light-protocol/pull/600/files/2c2c17017cd05c8fc82d77cfc55edcebe4cf5ee4#diff-c39d39cf374913058da43546ec23bde4639faf0e7619cd897823d21cff6d5bb9L45-L46) cmds, it can be confusing.
But it allows us to start server with only needed circuits => optimise CI execution.

.github/workflows/rust.yml
```diff
@@ -42,19 +42,26 @@ jobs:
...
+          source ./scripts/devenv.sh
+          npx nx build @lightprotocol/programs
 
-      - name: Build and test
+      - name: Build CLI
         run: |
           source ./scripts/devenv.sh
-          cargo build --workspace --all-targets
-          RUST_MIN_STACK=8388608 cargo test --workspace --all-targets
+          npx nx build @lightprotocol/cli
```
ananas-block, line 62: why do we need to build the cli to run rust tests?

circuit-lib/circuitlib-rs/src/gnark/non_inclusion_json_formatter.rs
```diff
@@ -1,62 +1,63 @@
...
-    values: Vec<String>,
-    leafLowerRangeValues: Vec<String>,
-    leafHigherRangeValues: Vec<String>,
-    leafIndices: Vec<u32>,
-    inPathIndices: Vec<u32>,
-    inPathElements: Vec<Vec<String>>,
+    root: String,
+    value: String,
+
+    #[serde(rename(serialize = "pathIndex"))]
```
ananas-block, line 23: nice

cli/src/commands/start-prover/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    process.exit();
+  }
+
+  static flags = {
+    "prove-compressed-accounts": Flags.boolean({
+      description: "Enable proving of compressed accounts.",
+      default: false,
+      char: "c",
+    }),
+    "prove-new-addresses": Flags.boolean({
```
ananas-block, line 18: can we by default start with all proving keys and add flags to skip provingkeys?
I feel like that way it will be easier to use.
``skip-prove-new-addresses``

cli/src/utils/proverUtils.ts
```diff
@@ -1,4 +1,4 @@
-import { getProverNameByArch } from "./initTestEnv";
+import { getProverNameByArch } from "./process";
```
ananas-block, line 1: can we rename ``process.ts`` to something more meaningful?
I am also seeing that ``process.ts`` combines functionality to start the test-validator and the prover server.
To get a more meaningful name maybe it makes sense to split the functionality into two files like:
1. ``process-start-test-validator.ts``
2. ``process-start-prover-server.ts`` 

or move the prover server part into `` prover-utils.ts``  and rename only ``process.ts`` to ``process-start-test-validator.ts``

ananas-block (commented):
Thanks for this pr it makes the code a lot simpler!

circuit-lib/circuitlib-rs/src/gnark/helpers.rs
```diff
@@ -29,42 +28,44 @@ impl Display for ProofType {
...
+    Command::new("sh")
+        .arg("-c")
+        .arg(cmd)
+        .spawn()
+        .expect("Failed to start server process");
+    prover_init_wait(cmd).await;
+}
+
+pub async fn spawn_prover() {
+    let cmd = "../../cli/test_bin/run start-prover -c -n";
```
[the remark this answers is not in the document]
sergeytimoshin: Yes, I wanted to unify the methods for spawning external programs, don't want to produce same logic in rust and in the cli.

PR #601, opened 2024-04-13
feat: SPL compression, decompression, refactors

SwenSchaeferjohann:
* refactored: JS c-token transfer and ix packing
* added: compress,decompress SPL ixs, actions, cli commands, test coverage (happy paths)
* adjusted programs to support compression, token, `mint` field into `create_transfer_instruction`() to support zero-token-inputs (e.g. for spl-compression)

note: locally we had to run with `cargo test-sbf -- --test-threads=1` otherwise we'd get gnark-server contention.


programs/compressed-token/src/process_transfer.rs
```diff
@@ -33,16 +34,21 @@ pub fn process_transfer<'a, 'b, 'c, 'info: 'b + 'c>(
             &ctx.accounts.authority.key(),
         )?;
 
+    // TODO: check if this is already implemented
```
SwenSchaeferjohann, line 37: this was a remnant from rebasing spl-compression onto the latest main: the delegation check is already implemented via `get_input_compressed_accounts_with_merkle_context_and_check_signer` correct? @ananas-block

programs/compressed-token/src/process_transfer.rs
```diff
@@ -456,9 +479,13 @@ pub mod transfer_sdk {
         leaf_indices: &[u32],
         proof: &CompressedProof,
         input_token_data: &[crate::TokenData],
+        mint: Pubkey,
```
SwenSchaeferjohann, line 482: FYI: we have to pass mint separately because in case of spl-compress, input_token_data is empty

programs/compressed-token/src/process_transfer.rs
ananas-block: Yes delegate checks are implemented but untested.
Also we don't have an instruction to create a compressed token account with a delegate yet.

programs/compressed-token/tests/test.rs
```diff
@@ -1,5 +1,7 @@
 #![cfg(feature = "test-sbf")]
 
+use std::{print, println};
```
ananas-block, line 3: why is this necessary?

programs/compressed-token/src/process_mint.rs
```diff
@@ -265,14 +268,15 @@ pub mod mint_sdk {
...
-    use crate::{get_token_authority_pda, get_token_pool_pda};
+    use crate::{get_cpi_authority_pda, get_token_authority_pda, get_token_pool_pda};
 
     pub fn create_initialize_mint_instruction(
         fee_payer: &Pubkey,
         authority: &Pubkey,
         mint: &Pubkey,
     ) -> Instruction {
         let token_pool_pda = get_token_pool_pda(mint);
+        println!("token_pool_pda: {:?}", token_pool_pda);
```
ananas-block, line 279: leftover

programs/compressed-token/src/process_transfer.rs
```diff
@@ -557,6 +584,8 @@ pub mod transfer_sdk {
             pubkey_array,
```
ananas-block, line 584: pubkey array has been deleted from this struct onchain but still exists here.
Happy to add the accounts to remaining accounts shouldn't be a big change.

js/compressed-token/src/actions/decompress.ts
```diff
@@ -0,0 +1,88 @@
...
+        outputStateTree: merkleTree,
+        recentInputStateRootIndices: proof.rootIndices,
+        recentValidityProof: proof.compressedProof,
+    });
+
+    const { blockhash } = await rpc.getLatestBlockhash();
+    const additionalSigners = dedupeSigner(payer, signers);
+    const signedTx = buildAndSignTx(ixs, payer, blockhash, additionalSigners);
+    const txId = await sendAndConfirmTx(rpc, signedTx, confirmOptions);
+    return txId;
```
ananas-block, line 87: on a general note, we could think about returning more information here such as which compressed accounts have been spent or created etc.

js/compressed-token/src/actions/transfer.ts
```diff
@@ -101,56 +58,70 @@ export async function transfer(
...
-        ),
-        [merkleTree, merkleTree], // out state trees
-        inputAccounts.map(account => account.compressedAccount), // input compressed accounts
-        [recipientCompressedAccount, changeCompressedAccount], // output compressed accounts
-        // TODO: replace with actual recent state root index!
-        // This will only work with sequential state updates and no cranking!
-        proof.rootIndices, // input state root indices
-        proof.compressedProof,
-    );
+    const ixs = await CompressedTokenProgram.transfer({
```
ananas-block, line 70: Naming is confusing to me ``createTransferInstruction`` is more intuitive in my opinion.
The name transfer leads me to expect for a transaction being sent not instructions being returned.

js/compressed-token/src/program.ts
```diff
@@ -4,27 +4,137 @@ import {
...
+    /**
+     * The payer of the transaction.
+     */
+    payer: PublicKey;
+    /**
+     * owner of the *uncompressed* token account.
+     */
+    owner: PublicKey;
+    /**
+     * source (associated) token account address.
```
ananas-block, line 49: actual address of the ata or the public key the ata is derived from?

js/compressed-token/src/program.ts
```diff
@@ -4,27 +4,137 @@ import {
...
+     * The recent state root indices of the input state. The expiry is tied to
+     * the proof.
+
+     */
+    recentInputStateRootIndices: number[];
+    /**
+     * The recent validity proof for state inclusion of the input state. It
+     * expires after n slots.
+     */
+    recentValidityProof: CompressedProof;
```
ananas-block, line 130: nice the recent makes a lot of sense, we should add an error for validity proof expired!

ananas-block: instead of the root not found error.

js/compressed-token/src/actions/decompress.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const proof = await rpc.getValidityProof(
+        inputAccounts.map(account => bn(account.compressedAccount.hash)),
+    );
+
+    const ixs = await CompressedTokenProgram.decompress({
+        payer: payer.publicKey,
+        inputCompressedTokenAccounts: inputAccounts,
+        toAddress: toAddress, // TODO: add check that it is a token account
+        amount: amount,
```
ananas-block, line 77:
```suggestion
        toAddress, // TODO: add check that it is a token account
        amount,
```

js/compressed-token/src/actions/compress.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    amount = bn(amount);
+
+    const ixs = await CompressedTokenProgram.compress({
+        payer: payer.publicKey,
+        owner: currentOwnerPublicKey,
+        source: sourceTokenAccount,
+        toAddress: toAddress,
+        amount: amount,
+        mint: mint,
```
ananas-block, line 62:
```suggestion
        toAddress,
        amount,
        mint,
```

js/compressed-token/src/actions/compress.ts
```diff
@@ -0,0 +1,71 @@
...
+    mint: PublicKey,
+    amount: number | BN,
+    owner: Signer | PublicKey,
+    sourceTokenAccount: PublicKey,
+    toAddress: PublicKey,
+    merkleTree: PublicKey = defaultTestStateTreeAccounts().merkleTree,
+    multiSigners: Signer[] = [],
+    confirmOptions?: ConfirmOptions,
+): Promise<TransactionSignature> {
+    const [currentOwnerPublicKey, signers] = getSigners(owner, multiSigners);
```
ananas-block, line 52: I am a bit confused by multiple signers and separate payer and owner, but need to think about how I would do it differently.

js/compressed-token/src/program.ts
```diff
@@ -130,7 +344,10 @@ export class CompressedTokenProgram {
     /** @internal */
     static get deriveCpiAuthorityPda(): PublicKey {
         const [address, _] = PublicKey.findProgramAddressSync(
-            [Buffer.from('cpi_authority')],
+            [
+                Buffer.from('cpi_authority'),
```
ananas-block, line 348: what about creating a constant for this or even taking the constant from the idl?

ananas-block (commented):
Awesome, looks good in general.
Couple small comments.

programs/compressed-token/tests/test.rs
SwenSchaeferjohann: removed unnecessary prints

js/compressed-token/src/program.ts
SwenSchaeferjohann: yea we can add a method that explicitly validates the recency, but i think `root not found` is pretty decent for a start. (equivalent to `blockhash not found`); client can reason that it's most likely bc expired.

js/compressed-token/src/program.ts
SwenSchaeferjohann: made constant

js/compressed-token/src/actions/compress.ts
SwenSchaeferjohann: yea multiSigners was a leftover from dealing with multisigs, (i.e. `mintAuthority` and `owner` being a multisig)
we'll have to deal with that in a future pr - straightforward support for multisigs.

js/compressed-token/src/actions/transfer.ts
SwenSchaeferjohann: yea agree it's more descriptive however .transfer() matches the solana systemprogram implementation, so there's familiarity for folks who work with it e.g. in browser environments...

js/compressed-token/src/actions/decompress.ts
SwenSchaeferjohann: yeah good idea, noted for a followup pr

vadorovsky (approved):

PR #602, opened 2024-04-13
chore: tmp: spl compression on-chain

SwenSchaeferjohann:
- wip. will be integrated into #601


SwenSchaeferjohann:
this is included in #601 , closing

PR #612, opened 2024-04-19
feat: cross-platform cli prover + unbundled photon

sergeytimoshin:
• Added cross-compilation step (arm64 & x64 for windows, linux, mac) for gnark-prover cli bundling
• Photon indexer removed from cli/bin, check at runtime if it is installed, otherwise throw error.


js/stateless.js/src/idls/index.ts
```diff
@@ -9,7 +9,10 @@ import {
     IDL as LightCompressedPdaIDL,
     LightCompressedPda,
 } from './light_compressed_pda';
-import { IDL as LightUserRegistryIDL, LightUserRegistry } from './light_user_registry';
+import {
+    IDL as LightUserRegistryIDL,
+    LightUserRegistry,
+} from './light_user_registry';
```
SwenSchaeferjohann, line 15: in line 7 of same file, we need to change 
`import { IDL as LightSystemIDL, LightRegistry } from './light_registry';`
to 
`import { IDL as LightRegistryIdl, LightRegistry } from './light_registry';`
as well.

vadorovsky (approved):

PR #620, opened 2024-04-22
refactor: rename indexed_array to nullifier_queue

sergeytimoshin:


vadorovsky (commented):
Looks good, but I think you need to regenerate accounts (`./scripts/getAccountState.sh`) to make the js tests pass.

ananas-block (approved):

PR #641, opened 2024-04-28
refactor: token transfer test

ananas-block:
Changes:
- refactored transfer test to test:
  - 1 in 1..11 outputs
  - 2 in 1..11 outputs
  - 3 in 1..11 outputs
  - 4 in 1..11 outputs
  - 8 in 1..8 outputs
- removed onchain prints

### DIFF
--- a/programs/compressed-token/tests/test.rs
+++ b/programs/compressed-token/tests/test.rs
@@ -17,8 +17,7 @@ use light_circuitlib_rs::{
 };
 use light_compressed_pda::CompressedProof;
 use light_compressed_pda::{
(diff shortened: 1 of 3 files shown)


programs/compressed-token/tests/test.rs
```diff
@@ -293,7 +311,7 @@ async fn test_transfer() {
         Some(TransactionParams {
             num_new_addresses: 0,
             num_input_compressed_accounts: 0,
-            num_output_compressed_accounts: 1,
+            num_output_compressed_accounts: inputs as u8,
```
SwenSchaeferjohann, line 314: why do we have to type cast as u8?

SwenSchaeferjohann (approved):

PR #658, opened 2024-05-02
feat: compressed token program - registerSpl, approveAndMintTo 

SwenSchaeferjohann:
* support for existing SPL mints. 
* we need a dedicated method for these because existing mints have external (relative to our system) mint authorities, therefore our system cannot directly mintTo compressed accounts. 

- [x] typedoc and cleanup compressed token program
- [x] instructions
- [x] actions
- [x] tests
- [x] cli commands
- [x] cli tests


cli/src/commands/approve-and-mint-to/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      required: true,
+    }),
+    to: Flags.string({
+      description: "Specify the recipient address.",
+      required: true,
+    }),
+    amount: Flags.integer({
+      description: "Amount to mint, in tokens.",
+      required: true,
+    }),
```
ananas-block, line 38: for required flags wouldn't it make sense to make these arguments?

js/compressed-token/src/types.ts
```diff
@@ -2,54 +2,115 @@ import { PublicKey } from '@solana/web3.js';
...
     lamports: BN | null;
 };
 
 export type InputTokenDataWithContext = {
+    /**
+     * The amount of tokens to transfer
+     */
     amount: BN;
+    /**
+     * Optional: The index of the delegate in the merkle tree
```
ananas-block, line 26: ``delegate in the merkle tree`` what does this mean?

ananas-block: is it the index in remaining accounts of the compressed tx?

ananas-block (approved):
looks good overall, no need to swap any flags for args right away but it feels more intuitive to me.
Also see my other comment.

cli/src/commands/approve-and-mint-to/index.ts
SwenSchaeferjohann: on bucket list for separate pr

PR #684, opened 2024-05-07
fix: concurrent merkle tree, wrong upper_leaf in update_proof_from_changelog  

sergeytimoshin:


merkle-tree/concurrent/tests/tests.rs
```diff
@@ -1212,3 +1215,125 @@ fn test_changelog_event_v1() {
...
+            .unwrap();
+        crank_merkle_tree
+            .append(&leaf_after_transfer_2.last().unwrap())
+            .unwrap();
+        // we should have a queue with 2 elements leaf payer leaf_after_transfer[0]
+        let nullifier_1 = BigUint::from_be_bytes(leaf_payer.as_slice());
+        let nullifier_2 = BigUint::from_be_bytes(leaf_after_transfer.first().unwrap());
+        let mut queue = HashSet::<u16>::new(6857, 4800, 2400).unwrap();
+        queue.insert(&nullifier_1, 1).unwrap();
+        queue.insert(&nullifier_2, 1).unwrap();
```
ananas-block, line 1292: can you remove the ugly hardcoded leaves?
(the test should work the same way without.)

merkle-tree/concurrent/src/changelog.rs
```diff
@@ -74,9 +74,9 @@ impl<const HEIGHT: usize> ChangelogEntry<HEIGHT> {
             // to retry the update with a new proof.
             //
             // TODO(vadorovsky): Re-visit optional throwing of this error.
-            // if !allow_updates_in_changelog {
-            //     return Err(ConcurrentMerkleTreeError::CannotUpdateLeaf);
-            // }
+            if !allow_updates_in_changelog {
+                return Err(ConcurrentMerkleTreeError::CannotUpdateLeaf);
+            }
```
ananas-block, line 79: you can remove this error @vadorovsky is removing it in his error pr right now

ananas-block: ( this is also what's failing the tests right now)

ananas-block (commented):
Thanks!
Can you add this  change as well?
https://github.com/Lightprotocol/light-protocol/blob/c454ebcb60f33d9289de0d1084e0b598855e1e3d/merkle-tree/concurrent/src/lib.rs#L896
It doesn't try to update the proof when the root is the same as onchain.

ananas-block:
weird that onchain tests are not triggered by rust library changes.
@sergeytimoshin can you enable program tests for changes on the merkle-tree/* folder as well?

vadorovsky (approved):

ananas-block (approved):

PR #702, opened 2024-05-11
feat: add option to pass custom photon db url to cli/test-validator

sergeytimoshin:


SwenSchaeferjohann:
needs rebase otherwise lgtm! @sergeytimoshin

SwenSchaeferjohann:
curious if there's a good way to test the use of a custom db properly in ci?

SwenSchaeferjohann (approved):

sergeytimoshin:
> curious if there's a simple way to test the use of a custom db properly in ci?

We don't have `test-validator` test setup yet, so the first logical step would be to create validator cmd test.
As regards to custom db tests, I'm using psql docker env for my local tests, so we can just reuse that: run `light test-validator` => then `pg_isready -- <db-url>`

PR #704, opened 2024-05-12
chore: esm node example, build cleanup

SwenSchaeferjohann:
* removes outdated web app
* bumps to latest web3js version
* web3js into peer-dependencies
* anchor into direct dependencies to enforce 0.29.0 while allowing consumers to use 0.30 (0.30 breaks stateless.js)
* general dependency tree cleanup
* test-helpers don't require hasher.rs, hasher -> devDependencies
* optionally, hasher.rs can be used directly by 3rd parties in their tests
* this prevents blowing up the default bundle size for consumers


js/stateless.js/src/test-helpers/index.ts
```diff
@@ -1,3 +1,3 @@
-/// TODO: remove and use test-helper pkg after photon is debugged.
+import BN from 'bn.js';
```
ananas-block, line 1: usually we use bn from anchor is there a reason why to do it differently here?

ananas-block (approved):

PR #707, opened 2024-05-13
chore: system program

SwenSchaeferjohann:
dir
/system/
crate
/light-system-program
import
LightSystemProgram, light_system_program


js/stateless.js/src/idls/light_compressed_pda.ts
```diff
@@ -1,6 +1,6 @@
-export type LightCompressedPda = {
+export type LightSystemProgram = {
     version: '0.3.1';
-    name: 'light_compressed_pda';
+    name: 'light_system_program';
```
ananas-block, line 3: this file should be removed right?

ananas-block (approved):

PR #716, opened 2024-05-14
refactor: update nullify_leaves instruction to emit nullify events

sergeytimoshin:


SwenSchaeferjohann:
Will this introduce a breaking change to downstream indexing?

ananas-block:
thank looks good to me, let's hold off until the releases are done.
Also it would make sense to add a unit test that the event is emitted correctly I can do that as well.

ananas-block (approved):

PR #745, opened 2024-05-15
test: add e2e svm test

ananas-block:
Changes:
- add e2e randomized svm test
- the goal is to add a short version with for example 100 iterations to CI
- and keep one longer version that we can run locally with ``#[ignore]`` in the code base
- removed StateMerkleTree type it was not used
- add address Merkle tree crank functionality
- add to CI
- assert new address are in address queue
- assert empty address queue and address Merkle tree update
- assert public transaction event after every transaction

Different pr:
[ ] rollover test


vadorovsky (commented):
(not fully done yet)

ananas-block:
> (not fully done yet)

what do you think of the overall design?

test-utils/src/test_forester.rs
```diff
@@ -0,0 +1,161 @@
...
+use solana_program_test::ProgramTestContext;
+use solana_sdk::signature::{Keypair, Signer};
+
+use crate::{
+    create_and_send_transaction, get_hash_set,
+    test_indexer::{AddressMerkleTreeAccounts, StateMerkleTreeAccounts, StateMerkleTreeBundle},
+    AccountZeroCopy,
+};
+
+///
```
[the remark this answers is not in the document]
ananas-block: turns out I don't need this struct at all I am removing it and am just using the function below

test-programs/rnd-inf-test/tests/test.rs
```diff
@@ -0,0 +1,884 @@
...
+            .position(|u| &u.keypair.pubkey() == user)
+            .unwrap();
+        // compress spl
+        // check sufficient spl balance
+        if self
+            .rng
+            .gen_bool(self.keypair_action_config.decompress_spl.unwrap_or(0.0))
+            && self.users[user_index].token_accounts.len() != 0
+        {
+            println!("\n --------------------------------------------------\n\t\t Compress Spl\n --------------------------------------------------");
```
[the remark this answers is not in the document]
ananas-block: then it doesn't stand out in the logs svm logs are really noisy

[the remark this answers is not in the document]
vadorovsky: I see.

Now there is only one big `println!` like that, so I think it's fine. I think that yesterday I've seen multiple logs doing that and it looked a bit untidy.

If we ever want to have more prints like that, I think it would be worth moving to some function or macro, so we don't keep repeating >200 char lines.

vadorovsky:
> > (not fully done yet)
> 
> what do you think of the overall design?

Overall I like it. I like how some bits of already existing tests are being re-used here and the fact you can parametrize `InfTestEnv`. I hope it will push us more and more to have particular test-cases as importable functions, which are going to be used across the regular and randomized tests.

I'm only wondering - do you think it would make sense to have such randomized tests also for more low-level operations? The current tests check mostly compression/decompression. Would be nice to leverage this framework also to leaf appends, adding addresses etc. as single operations, to make sure that these instructions work fine. I know users won't interact with these much, but still.

merkle-tree/indexed/tests/tests.rs
```diff
@@ -540,113 +540,115 @@ pub fn functional_non_inclusion_test() {
...
+// TODO: figure out address Merkle tree changelog
+// #[test]
+// pub fn functional_changelog_test() {
+//     let address_1 = 30_u32.to_biguint().unwrap();
+//     let address_2 = 10_u32.to_biguint().unwrap();
+
+//     perform_change_log_test(address_1.clone(), address_2.clone());
+// }
+
+// fn perform_change_log_test(address_1: BigUint, address_2: BigUint) {
```
vadorovsky, line 569: Is this test commented because of the broken concurrency problems you talked about? I'm happy to help tomorrow with debugging it, if it's still a blocker.

test-utils/src/address_tree_rollover.rs
```diff
@@ -0,0 +1,356 @@
...
+
+pub async fn set_address_merkle_tree_next_index(
+    context: &mut ProgramTestContext,
+    merkle_tree_pubkey: &Pubkey,
+    next_index: u64,
+    lamports: u64,
+) {
+    // is in range 8 -9 in concurrent mt
+    // offset for next index
+    let offset_start = 6 * 8 + 8 + 4 * 32 + 8 * 8;
```
vadorovsky, line 30: memoffset, pls :)

Also, a note for myself mostly - I really need to make the "copy and deserialize account" abstractions less painful and Tokio-compatible. I can totally understand why do you prefer to get data by offsets, but the goal would be to make them robust enough so nobody thinks of doing this. ^

test-utils/src/state_tree_rollover.rs
```diff
@@ -0,0 +1,377 @@
...
+    context: &mut ProgramTestContext,
+    merkle_tree_pubkey: &Pubkey,
+    next_index: u64,
+    lamports: u64,
+) {
+    // is in range 8 -9 in concurrent mt
+    // offset for next index
+
+    let offset_start = 8
+        + offset_of!(StateMerkleTreeAccount, state_merkle_tree_struct)
```
vadorovsky, line 115: Like here, nice :+1:

test-programs/rnd-inf-test/tests/test.rs
```diff
@@ -0,0 +1,932 @@
...
+    }
+
+    /// 1. Transfer spl tokens between random users
+    pub async fn activate_keypair_actions(&mut self, user: &Pubkey) {
+        let user_index = self
+            .users
+            .iter()
+            .position(|u| &u.keypair.pubkey() == user)
+            .unwrap();
+        // compress spl
```
vadorovsky, line 419: Do you think we could move each operation - `compress spl`, `decompress spl`, `transfer spl`, `create address`, `compress sol`, `decompress sol`, `transfer sol` - to separate methods? From a brief look I can't see any of them sharing too much variables. And it would make easier for others to figure out the sequence of actions.

test-programs/rnd-inf-test/tests/test.rs
```diff
@@ -0,0 +1,932 @@
...
+        {
+            println!("\n --------------------------------------------------\n\t\t Compress Spl\n --------------------------------------------------");
+            let mut balance = 0;
+            let mut mint = Pubkey::default();
+            let mut token_account = Pubkey::default();
+            for _ in 0..self.users[user_index].token_accounts.len() {
+                let (_mint, _token_account) = self.users[user_index].token_accounts[self
+                    .rng
+                    .gen_range(0..self.users[user_index].token_accounts.len())];
+                token_account = _token_account;
```
vadorovsky, line 434: Just wondering - do you expect to have the token accounts also somehow randomized?

So far, from my understanding, this loop will always take the first token account with positive balance.

I see that later, we are using this utility to pick a randomized SPL token account

```rust
let (mint, token_accounts) = self.select_random_spl_token_accounts(user).await;
```

Would that be applicable here, and in other places using the 1st non-empty token account, as well? I see that `self.select_random_spl_token_accounts` already filters for non-empty.

ananas-block: > do you expect to have the token accounts also somehow randomized?
every iteration selects a token account randomly by generating a random index with:
`` self.rng.gen_range(0..self.users[user_index].token_accounts.len())``
is this what you mean?

yeah it could be a method

ananas-block: wrt ``select_random_spl_token_accounts``
naming is off it should be named ``select_random_compressed_token_accounts``

merkle-tree/indexed/tests/tests.rs
ananas-block: This code is commented because the concurrency feature was breaking my test.
It's not directly a blocker my tests run since I removed concurrency.
But we don't have address Merkle tree concurrency.
We definitely need address Merkle tree concurrency for mainnet.
But for this audit I want to get everything else working before looking into concurrency again.

ananas-block:
> do you think it would make sense to have such randomized tests also for more low-level operations?

Many lowlevel operations are covered implicitly and asserted explicity:
- append is part of every transaction with output compressed accounts
  (we assert that the Merkle tree contains the hashes of output compressed accounts)
- inserts into queues is part of every tx with input compressed accounts
  (we assert that the queue contains the nullified compressed accounts)
  
Other lowlevel operations are being tested explicitly:
- nullify
- insert address (when emptying the address queue)

In general the more tests the better but timewise we need to prioritze.

vadorovsky (approved):

PR #759, opened 2024-05-21
refactor: Wrap ProgramTestContext into the RpcConnection trait

sergeytimoshin:


test-programs/account-compression-test/tests/address_merkle_tree_tests.rs
```diff
@@ -187,7 +202,7 @@ async fn test_insert_invalid_low_element() {
         bigint_to_be_bytes_array(&address1).unwrap(),
         bigint_to_be_bytes_array(&address2).unwrap(),
     ];
-    insert_addresses(
+    let _ = insert_addresses(
```
ananas-block, line 205: would it make sense to unwrap twice instead of putting a result or option into an placeholder variable?

sergeytimoshin: Yes, thanks, I cleaned it up.

test-utils/src/rpc/rpc_connection.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn set_account(&mut self, address: &Pubkey, account: &AccountSharedData);
+    fn get_rent(&mut self) -> impl std::future::Future<Output = Result<Rent, RpcError>> + Send;
+    fn get_latest_blockhash(
+        &mut self,
+    ) -> impl std::future::Future<Output = Result<Hash, RpcError>> + Send;
+    fn process_transaction_with_metadata(
+        &mut self,
+        transaction: Transaction,
+    ) -> impl std::future::Future<Output = Result<(), RpcError>> + Send;
+    fn get_root_slot(&mut self) -> impl std::future::Future<Output = Result<u64, RpcError>> + Send;
```
ananas-block, line 43: in rpc this is just get_slot.
https://docs.rs/solana-client/latest/solana_client/nonblocking/rpc_client/struct.RpcClient.html#method.get_slot

test-utils/src/rpc/rpc_connection.rs
```diff
@@ -0,0 +1,59 @@
...
+    ) -> impl std::future::Future<Output = Result<Option<Account>, RpcError>> + Send;
+    fn set_account(&mut self, address: &Pubkey, account: &AccountSharedData);
+    fn get_rent(&mut self) -> impl std::future::Future<Output = Result<Rent, RpcError>> + Send;
+    fn get_latest_blockhash(
+        &mut self,
+    ) -> impl std::future::Future<Output = Result<Hash, RpcError>> + Send;
+    fn process_transaction_with_metadata(
+        &mut self,
+        transaction: Transaction,
+    ) -> impl std::future::Future<Output = Result<(), RpcError>> + Send;
```
ananas-block, line 42: in program test this actually returns meta data.
What about changing it to process_transaction?
``
    fn send_transaction(
        &mut self,
        transaction: Transaction,
    ) -> impl std::future::Future<Output = Result<Signature, RpcError>> + Send;
    ``

test-utils/src/rpc/test_rpc.rs
```diff
@@ -0,0 +1,254 @@
...
+        // Send the transaction
+        self.context
+            .banks_client
+            .process_transaction(transaction)
+            .await?;
+
+        Ok(())
+    }
+
+    // TODO: return Result<T, Error>
```
ananas-block, line 237: we should move this todo to the trait

test-utils/src/rpc/rpc_connection.rs
ananas-block: the equivalent in solana client
https://docs.rs/solana-client/latest/solana_client/nonblocking/rpc_client/struct.RpcClient.html#method.send_and_confirm_transaction

ananas-block (commented):
Apart from the comments looks good, thanks!

test-utils/src/rpc/rpc_connection.rs
sergeytimoshin: 👍

test-utils/src/rpc/rpc_connection.rs
sergeytimoshin: I renamed  `process_transaction_with_metadata` to `process_transaction` + we already have 
 ```
 fn create_and_send_transaction(
        &mut self,
        instruction: &[Instruction],
        payer: &Pubkey,
        signers: &[&Keypair],
    ) -> impl std::future::Future<Output = Result<Signature, RpcError>> + Send;
```

test-utils/src/rpc/test_rpc.rs
sergeytimoshin: 👍

ananas-block (approved):

PR #766, opened 2024-05-23
feat: combined proofs and validityProof refactor in rpc-interface

SwenSchaeferjohann:
* refactors test-rpc
* combined proofs as part of getValidityProof with unified interface
* added test coverage for account creation with compressed lamports (using combined proofs)
* removes an address print on-chain

### DIFF
--- a/js/stateless.js/src/actions/create-account.ts
+++ b/js/stateless.js/src/actions/create-account.ts
@@ -127,12 +127,10 @@ export async function createAccountWithLamports(
         payer.publicKey,
     );
 
-    console.log('ca1: ', compressedAccounts);
     const [inputAccounts] = selectMinCompressedSolAccountsForTransfer(
         compressedAccounts,
         lamports,
     );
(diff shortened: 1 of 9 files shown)


js/stateless.js/src/actions/create-account.ts
```diff
@@ -84,3 +89,93 @@ export async function createAccount(
...
+
+    const proof = await rpc.getValidityProof(
+        inputAccounts.map(account => bn(account.hash)),
+        [bn(address.toBytes())],
+    );
+
+    const params: NewAddressParams = {
+        seed: seed,
+        addressMerkleTreeRootIndex:
+            proof.rootIndices[proof.rootIndices.length - 1], // TODO: confirm
```
ananas-block, line 151: in rust we have address Merkle tree rootIndices and state Merkle tree rootIndices.
How do you distinguish these in ts?

ananas-block (approved):

PR #771, opened 2024-05-28
feat: public transaction event add seq

ananas-block:
Issue:
- we removed changelog events
- public transaction events contain the input compressed account hashes but no sequence number

Changes:
- add sequence number for every output state Merkle tree to public transaction event


SwenSchaeferjohann (approved):

SwenSchaeferjohann (approved):
thx!

PR #772, opened 2024-05-28
chore: rm mintAuthorityPda as proxy signer for mintTo

SwenSchaeferjohann:
### Changes 

compressed-token program:

* `create-mint`: 
    * Remove any `authority` signer account. `Create-mint` registers the mint by creating the `tokenPoolPda`
    * `tokenPoolPda` is derived from `Mint`, not from `authority`.
    * Anyone can create this PDA, which allows any account to compress tokens of said mint.
        * This is required so that mints with burned `mintAuthority` can still be compressed.

* `Mint-to`
    * `process_mint_to`
        * Removed `mintAuthorityPda`, instead...  
        * Use static `cpiAuthority` to CPI into the system program. 
        * No `mintAuthority` check in our system is required because `mint_spl_to_pool_pda` already does the `mintAuthority` signer check
            * CPI into SPL mint-to passes authority (`mintAuthority`), which has to sign the tx

* Added `get_cpi_signer_seeds`() helper because its reused in process_transfer
* Removed `get_token_authority_pda` because it is unused now

All this simplifies the JS client and CLI:

* `createMint` can now reuse `registerMint` logic because authorities are dealt with consistently.
* `mintAuthority` is no longer required as TX signer for `register-mint` (this was prev. unchecked), and `create-mint`. So we pass it as pubkeys now.

### DIFF
--- a/programs/compressed-token/src/process_mint.rs
+++ b/programs/compressed-token/src/process_mint.rs
@@ -1,8 +1,9 @@
-use crate::{
-    constants::TOKEN_COMPRESSED_ACCOUNT_DISCRIMINATOR, token_data::{AccountState, TokenData}
-};
 #[cfg(target_os = "solana")]
 use crate::get_cpi_signer_seeds;
+use crate::{
+    constants::TOKEN_COMPRESSED_ACCOUNT_DISCRIMINATOR,
+    token_data::{AccountState, TokenData},
+};
 use anchor_lang::prelude::*;
 use anchor_spl::token::{Mint, Token, TokenAccount};
 use light_hasher::Poseidon;
@@ -22,8 +23,6 @@ pub const POOL_SEED: &[u8] = b"pool";
 pub struct CreateMintInstruction<'info> {
     #[account(mut)]
     pub fee_payer: Signer<'info>,
-    #[account(mut)]
-    pub authority: Signer<'info>,
     #[account(init,
               seeds = [
                 POOL_SEED, &mint.key().to_bytes(),
@@ -49,7 +48,6 @@ pub fn process_mint_to<'info>(
     ctx: Context<'_, '_, '_, 'info, MintToInstruction<'info>>,
     compression_public_keys: Vec<Pubkey>,
     amounts: Vec<u64>,
-    // bump: u8,
 ) -> Result<()> {
     if compression_public_keys.len() != amounts.len() {
         msg!(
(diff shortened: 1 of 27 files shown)


programs/compressed-token/src/process_mint.rs
```diff
@@ -527,30 +525,31 @@ fn test_manual_ix_data_serialization_borsh_compat() {
...
-    let authority_bytes = Pubkey::new_unique().to_bytes();
-    let mint_bytes = Pubkey::new_unique().to_bytes();
-    let bump = &[255];
-    let seeds = [
-        MINT_AUTHORITY_SEED,
-        authority_bytes.as_slice(),
-        mint_bytes.as_slice(),
-        bump,
-    ];
-    let inputs_struct = light_system_program::InstructionDataInvokeCpi {
```
ananas-block, line 539: we still need to call the light system program with invoke_cpi.

programs/compressed-token/src/process_mint.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(target_os = "solana")]
```
ananas-block, line 1: why is this necessary?

programs/compressed-token/src/process_mint.rs
```diff
@@ -360,12 +340,13 @@ pub struct MintToInstruction<'info> {
     pub fee_payer: Signer<'info>,
     #[account(mut)]
     pub authority: Signer<'info>,
-    // This is the cpi signer
+    // This is the cpi signer.
+    // TODO: double check security
```
ananas-block, line 344: we can remove the double check comment

programs/compressed-token/src/process_mint.rs
```diff
@@ -360,12 +340,13 @@ pub struct MintToInstruction<'info> {
...
     /// CHECK: that mint authority is derived from signer
-    #[account(mut, seeds = [MINT_AUTHORITY_SEED, authority.key().to_bytes().as_slice(), mint.key().to_bytes().as_slice()], bump,)]
-    pub mint_authority_pda: UncheckedAccount<'info>,
+    #[account(seeds = [b"cpi_authority"], bump,)]
```
ananas-block, line 346: don't we have a constant for this?

programs/compressed-token/src/process_transfer.rs
```diff
@@ -191,6 +191,14 @@ pub fn add_token_data_to_input_compressed_accounts(
     Ok(())
 }
 
+// TODO: consider moving this function to helpers
+/// Get static cpi signer seeds
+pub fn get_cpi_signer_seeds() -> [&'static [u8]; 2] {
+    let bump: &[u8; 1] = &[255];
+    let seeds: [&'static [u8]; 2] = [b"cpi_authority", bump];
```
ananas-block, line 198:
```suggestion
    let seeds: [&'static [u8]; 2] = [CPI_AUTHORITY_SEED, bump];
```

programs/compressed-token/src/process_transfer.rs
```diff
@@ -201,10 +209,10 @@ pub fn cpi_execute_compressed_transaction_transfer<'info>(
 ) -> Result<()> {
     bench_sbf_start!("t_cpi_prep");
 
-    let bump = &[255];
-    let seeds: [&[u8]; 2] = [b"cpi_authority".as_slice(), bump];
+    let signer_seeds = get_cpi_signer_seeds();
+    let signer_seeds_ref = &[&signer_seeds[..]];
 
-    let signer_seeds = &[&seeds[..]];
+    // let signer_seeds = &[&seeds[..]];
```
ananas-block, line 215: leftover?

ananas-block (commented):
thanks!

programs/compressed-token/src/process_transfer.rs
SwenSchaeferjohann: will do in follow-up pr with all other occurrences

programs/compressed-token/src/process_mint.rs
SwenSchaeferjohann: will do in follow-up pr with all other occurrences

programs/compressed-token/src/process_mint.rs
SwenSchaeferjohann: cargo clippy throws unused import otherwise because `get_cpi_signer_seeds` is only used in target_os = "solana"

ananas-block (approved):

PR #775, opened 2024-05-30
chore: make outputStateTrees optional across all JS bindings

SwenSchaeferjohann:
* this improves dx considerably for anyone who doesn't want to have to deal with picking output state trees explicitly, while still building txs themselves.


SwenSchaeferjohann:
ready for review, but must be rebased if merged after #772

vadorovsky (approved):

PR #776, opened 2024-05-30
test: system program cpi

ananas-block:
Refactors:
- cpi signers check should throw a different error (Signer seeds did not derive the correct signer)
- can write into a program owned Merkle tree
- can invalidate compressed accounts from a program owned Merkle tree


SwenSchaeferjohann (commented):
lgtm. just one test failed with https://github.com/Lightprotocol/light-protocol/actions/runs/9307209493/job/25619133995?pr=776

SwenSchaeferjohann (approved):

PR #785, opened 2024-06-01
chore: check rollover fee is high enough

ananas-block:
Changes:
- add a test to compute and assert that our current rollover fees cover the rentexemption of queue, Merkle tree and cpi context accounts for state and address accounts


SwenSchaeferjohann (commented):
lgtm! 
thoughts on adding?
* tests that are meant to fail
* tests around the boundary values
* any risks w.r.t. overflows / inprecision due to floating point math?

SwenSchaeferjohann:
checking lamports

SwenSchaeferjohann (approved):

PR #786, opened 2024-06-02
chore: review renamings, sum check tests

ananas-block:


vadorovsky (approved):
Looks good, but

* IDL needs regeneration
* there is one clippy error

PR #791, opened 2024-06-02
chore: feature gate cpi context

ananas-block:
chore(compressed-token): add sum failing test


programs/system/src/invoke_cpi/processor.rs
SwenSchaeferjohann, line 1: think we should feature gate in c-token program as well. bindings can stay the same either way, tho in cpi_execute_compressed_transaction_transfer we can gate this: 
   let cpi_context_account = cpi_context.map(|cpi_context| {
        ctx.remaining_accounts[cpi_context.cpi_context_account_index as usize].to_account_info()
    }); and just invokecpi to system program with None always

SwenSchaeferjohann (approved):

PR #805, opened 2024-06-06
feat: forester registration

ananas-block:
### Assumptions:
- only registered foresters can nullify or empty address queues
- foresters can only be registered by the governance authority

### Changes:
- EnvAccounts:
  - add forester: Keypair, the hardcoded test keypair of a registered forester
  - add registered_forester_epoch_pda: the register pda of the forester keypair
- regjstry:
  - add sdk create instruction functions for:
    - register forester
    - nullify
    - rollover
    - update address Merkle tree
  - add failing tests for functions that should only be served by registered foresters
- cli accounts:
  - add registered forester pda
  - modified governance struct hence 
- forester:
  - changed many RpcClient instantiations to confirmed commitment instead of finalized default
  - refactored empty address tree and nullify tests to only create addresses and produce nullifier queue entries instead of using random transactions (this makes the tests much faster)
- assert_rpc_error
  - simplied it by making the Ok() result generic


forester/src/nullifier/nullify.rs
```diff
@@ -84,7 +82,8 @@ pub async fn nullify<T: Indexer>(indexer: T, config: &Config) -> Result<(), Fore
             .compressed_account_proofs
             .remove(&account.hash_string())
         {
-            let client = RpcClient::new(&config.server_url);
+            let client =
+                RpcClient::new_with_commitment(&config.server_url, CommitmentConfig::confirmed());
```
SwenSchaeferjohann, line 86: what's the difference between RpcClient and RpcConnection?

SwenSchaeferjohann: think we should make use of the `get_rpc_client_confirmed` helper that you added below

forester/src/nullifier/nullify.rs
```diff
@@ -261,6 +245,7 @@ pub async fn nullify_compressed_account<T: Indexer>(
 
     let tx_config = RpcSendTransactionConfig {
         skip_preflight: true,
+        preflight_commitment: Some(solana_sdk::commitment_config::CommitmentLevel::Confirmed),
```
SwenSchaeferjohann, line 248: is this required? aren't we skipping preflight here?

forester/tests/empty_address_tree_test.rs
```diff
@@ -63,26 +50,31 @@ async fn empty_address_tree_test() {
...
+    }
+
     assert_ne!(get_state_queue_length(&config), 0);
     info!(
         "Address merkle tree: nullifying queue of {} accounts...",
         get_state_queue_length(&config)
     );
 
-    let mut rpc = SolanaRpcConnection::new(Some(CommitmentConfig::finalized())).await;
+    let mut rpc = SolanaRpcConnection::new(Some(CommitmentConfig::confirmed())).await;
```
SwenSchaeferjohann, line 68: let's also provide a wrapper for this? `get_solana_rpc_connection_confirmed`

test-utils/src/test_env.rs
```diff
@@ -126,6 +131,13 @@ pub const REGISTRY_ID_TEST_KEYPAIR: [u8; 64] = [
     20, 96, 192,
 ];
 
+pub const FORESTER_TEST_KEYPAIR: [u8; 64] = [
+    81, 4, 133, 152, 100, 67, 157, 52, 66, 70, 150, 214, 242, 90, 65, 199, 143, 192, 96, 172, 214,
+    44, 250, 77, 224, 55, 104, 35, 168, 1, 92, 200, 204, 184, 194, 21, 117, 231, 90, 62, 117, 179,
+    162, 181, 71, 36, 34, 47, 49, 195, 215, 90, 115, 3, 69, 74, 210, 75, 162, 191, 63, 51, 170,
+    204,
+];
```
SwenSchaeferjohann, line 139: can we move this to a central place?

SwenSchaeferjohann (approved):

PR #811, opened 2024-06-09
refactor: forester + tests

sergeytimoshin:
• use only one rpc instance in forester
• `create_and_send_transaction_with_event`  return signature along with event 
• add test helpers to test_env


test-utils/src/rpc/solana_rpc.rs
```diff
@@ -149,46 +149,46 @@ impl RpcConnection for SolanaRpcConnection {
...
+        //         post_balance as i64 - expected_post_balance
+        //     );
+        //     println!("network_fee: {}", network_fee);
+        //     return Err(RpcError::from(BanksClientError::TransactionError(
+        //         TransactionError::InstructionError(0, InstructionError::Custom(11111)),
+        //     )));
+        // }
+        // }
+
+        let result = event.map(|event| (event, signature));
```
SwenSchaeferjohann, line 190: think we'll need to enable these again for merging

forester/tests/test_indexer.rs
```diff
@@ -29,42 +34,257 @@ async fn test_indexer() {
...
         None,
         "../circuit-lib/circuitlib-rs/scripts/prover.sh",
     )
     .await;
     env.execute_rounds().await;
-    let indexer = env.indexer;
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn photon_interop_state_nullification_test() {
```
SwenSchaeferjohann, line 74: do we still need this test after #810  gets merged?

forester/src/nullifier/nullify.rs
```diff
@@ -41,7 +41,8 @@ pub async fn nullify<T: Indexer>(indexer: T, config: &Config) -> Result<(), Fore
 
     loop {
         let queue_data = {
-            let mut queue_data = fetch_queue_data(arc_indexer.clone(), config).await?;
+            let mut queue_data = fetch_queue_data(indexer, rpc, config).await?;
+            info!("Queue data: {:?}", queue_data);
```
ananas-block, line 45: This will log a lot of data.

forester/tests/nullifier_test.rs
```diff
@@ -32,8 +34,10 @@ fn test_config() -> Config {
 #[ignore]
 async fn queue_info_test() {
     let config = test_config();
-    let client = RpcClient::new(config.server_url);
-    let queue = get_nullifier_queue(&config.nullifier_queue_pubkey, &client).unwrap();
+    let mut rpc = SolanaRpcConnection::new(None).await;
```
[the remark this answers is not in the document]
sergeytimoshin: No, it's already there, right on the next line, we reuse this rpc connection through the forester internals:

 https://github.com/Lightprotocol/light-protocol/pull/811/files#diff-b1dc72fd951f332ca2b32620374dbe4f2c407c91a57c91943e43cdcd627b2a17R38

package.json
```diff
@@ -8,5 +8,6 @@
     "syncpack": "^11.2.1",
     "typescript": "5.3.2",
     "prettier": "^3.2.5"
-  }
+  },
+  "packageManager": "pnpm@9.2.0+sha512.98a80fd11c2e7096747762304106432b3ddc67dcf54b5a8c01c93f68a2cd5e05e6821849522a06fb76284d41a2660d5e334f2ee3bbf29183bf2e739b1dafa771"
```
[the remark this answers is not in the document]
sergeytimoshin: 100% not needed, weird leftover

forester/src/indexer/photon_indexer.rs
```diff
@@ -76,6 +77,37 @@ impl Indexer for PhotonIndexer {
...
+        };
+
+        let result = photon_api::apis::default_api::get_compressed_accounts_by_owner_post(
+            &self.configuration,
+            request,
+        )
+        .await
+        .unwrap();
+
+        info!("PhotonIndexer: Got response: {:?}", result);
```
SwenSchaeferjohann, line 100: leftover?

test-programs/e2e-test/tests/test.rs
```diff
@@ -1,28 +1,32 @@
...
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
 async fn test_10_validator_all() {
     spawn_test_validator().await;
     let env_accounts = get_test_env_accounts();
     let rpc = SolanaRpcConnection::new(None);
-
+    rpc.client
+        .request_airdrop(&rpc.get_payer().pubkey(), LAMPORTS_PER_SOL * 1000)
+        .unwrap();
+    tokio::time::sleep(std::time::Duration::from_secs(16)).await;
```
SwenSchaeferjohann, line 29: Iirc, @ananas-block discovered that we can use a random keypair here for the payer as a hotfix. The underlying issue is that the same hardcoded keypair gets used elsewhere in a completely unrelated test at the same time, and therefore, the airdrop here updates its balance during that other test run, causing that other test to fail.

pnpm-lock.yaml
SwenSchaeferjohann, line 1: any idea why this is happening?

test-programs/e2e-test/tests/test.rs
SwenSchaeferjohann, line 29: same here.

SwenSchaeferjohann (commented):
just these notes here - otherwise lgtm!

forester/tests/test_indexer.rs
```diff
@@ -29,6 +30,12 @@ async fn test_indexer() {
     };
 
     let rpc = SolanaRpcConnection::new(None);
+
+    rpc.client
+        .request_airdrop(&rpc.get_payer().pubkey(), LAMPORTS_PER_SOL * 1000)
+        .unwrap();
+    tokio::time::sleep(std::time::Duration::from_secs(16)).await;
```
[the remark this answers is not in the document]
ananas-block: agreed we should use confirm to wait for confirmation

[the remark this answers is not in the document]
sergeytimoshin: Thanks! Fixed.

pnpm-lock.yaml
sergeytimoshin: desperate attempt to fix weird ci cache issue by calling pnpm update

SwenSchaeferjohann (commented):
thx lgtm, any rem. thoughts @ananas-block ?

test-utils/src/spl.rs
```diff
@@ -59,7 +59,7 @@ pub async fn mint_tokens_helper<const INDEXED_ARRAY_SIZE: usize, R: RpcConnectio
         spl_token::state::Account::unpack(&rpc.get_account(pool).await.unwrap().unwrap().data)
             .unwrap()
             .amount;
-    let event = rpc
+    let (event, _signature) = rpc
```
ananas-block, line 62:
```suggestion
    let (event, _) = rpc
```
this works as well and produces less noise.

test-utils/src/rpc/test_rpc.rs
```diff
@@ -25,7 +25,7 @@ impl RpcConnection for ProgramTestRpcConnection {
         payer: &Pubkey,
         signers: &[&Keypair],
         transaction_params: Option<TransactionParams>,
-    ) -> Result<Option<T>, RpcError>
+    ) -> Result<Option<(T, solana_sdk::signature::Signature)>, RpcError>
```
ananas-block, line 28: A result struct would be a lot more intuitive than a tuple in my opinion.
Let's do that in a separate pr though.

test-utils/src/rpc/solana_rpc.rs
```diff
@@ -269,15 +265,18 @@ impl RpcConnection for SolanaRpcConnection {
...
+
+        loop {
+            let confirmed = self
+                .client
+                .confirm_transaction_with_commitment(&signature, self.client.commitment())?
+                .value;
+            if confirmed {
+                break;
+            }
         }
```
ananas-block, line 277: Also no need to do that in this pr but we need to get rid of these loops without a timeout.
If the transaction fails this will result in an infinite loop.

scripts/install.sh
```diff
@@ -110,7 +110,7 @@ check_flag() {
 
 GO_VERSION="1.21.7"
 NODE_VERSION="20.9.0"
-PNPM_VERSION="8.8.0"
+PNPM_VERSION="9.2.0"
```
ananas-block, line 113: could this cause the pnpm lock changes?

forester/tests/test_indexer.rs
```diff
@@ -50,20 +57,18 @@ async fn test_indexer() {
...
-    assert_ne!(get_state_queue_length(&config), 0);
+    assert_ne!(get_state_queue_length(&mut env.rpc, &config).await, 0);
     info!(
         "Nullifying queue of {} accounts...",
-        get_state_queue_length(&config)
+        get_state_queue_length(&mut env.rpc, &config).await
     );
-    nullify(indexer, &config).await.unwrap();
-    assert_eq!(get_state_queue_length(&config), 0);
+    let _ = nullify(&mut env.indexer, &mut env.rpc, &config).await;
```
ananas-block, line 65: why not use the result here?
We are awaiting the transaction so I don't see why we shouldn't fail if the transaction fails.

forester/tests/nullifier_test.rs
```diff
@@ -87,15 +91,15 @@ async fn tree_info_test() {
...
+    let mut indexer = PhotonIndexer::new(INDEXER_URL.to_string());
     let config = test_config();
-    let client = get_rpc_client_confirmed();
-    client
-        .request_airdrop(&config.payer_keypair.pubkey(), LAMPORTS_PER_SOL * 1000)
-        .unwrap();
+    let mut rpc = SolanaRpcConnection::new(None);
+    let _ = rpc
+        .airdrop_lamports(&config.payer_keypair.pubkey(), LAMPORTS_PER_SOL * 1000)
+        .await;
```
ananas-block, line 99: same here why not throw an error if the airdrop fails?

forester/tests/nullifier_test.rs
```diff
@@ -110,9 +114,9 @@ async fn test_nullify_leaves() {
...
     let config = test_config();
-    let client = get_rpc_client_confirmed();
-    client
-        .request_airdrop(&config.payer_keypair.pubkey(), LAMPORTS_PER_SOL * 1000)
-        .unwrap();
-    subscribe_nullify(&config).await;
+    let mut rpc = SolanaRpcConnection::new(None);
+    let _ = rpc
+        .airdrop_lamports(&config.payer_keypair.pubkey(), LAMPORTS_PER_SOL * 1000)
+        .await;
```
ananas-block, line 120: and here

forester/src/nullifier/nullify.rs
```diff
@@ -86,64 +92,55 @@ pub async fn nullify<T: Indexer>(indexer: T, config: &Config) -> Result<(), Fore
...
             let config_clone = config.clone();
-            let task = tokio::spawn(async move {
-                let _permit = permit;
-                let mut retries = 0;
-                loop {
-                    if cancellation_token_clone.is_cancelled() {
-                        info!("Task cancelled for account {}", account.hash_string());
+            let _permit = permit;
+            let mut retries = 0;
+            loop {
```
ananas-block, line 100: on testnet this retry logic will not work. It will hit the rpc ratelimit immediately.
You will need a delay for retries.

ananas-block: It probably makes sense to create a function send with retries.

ananas-block (commented):
Got a couple comments.
The only thing that I would want to discuss and potentially fix in this pr are the unused results.
(``let _ = function_that_returns_result().await;``)

scripts/install.sh
sergeytimoshin: No. pnpm-lock was updated when I called pnpm update.

test-utils/src/rpc/solana_rpc.rs
sergeytimoshin: Hm, as far as I understand, If the transaction fails, it will break the loop due to the error propagation through `?`. 
The only theoretical possibility of an infinite loop: transaction never confirmed.

Also, https://github.com/anza-xyz/agave/blob/master/rpc-client/src/rpc_client.rs#L979-L985

forester/src/nullifier/nullify.rs
sergeytimoshin: It has the delay here: https://github.com/Lightprotocol/light-protocol/pull/811/files/44d4ead2d6dc267dfbb4511305667fe0cd125aaa#diff-fed5e22bba6202e7a68545ef7f19570dbd3c084c5aa2f0d46703fd176a94371cR138

forester/tests/nullifier_test.rs
sergeytimoshin: Thanks, fixed!

forester/tests/nullifier_test.rs
sergeytimoshin: Fixed.

forester/tests/test_indexer.rs
sergeytimoshin: Fixed.

scripts/install.sh
ananas-block: ah I see

test-utils/src/rpc/solana_rpc.rs
ananas-block: I have been stuck in a loop like this when testing a couple days ago.
In the case where I was stuck, the rpc returned transaction signature even though tx has failed and I was able to look at the failed tx in the explorer.

ananas-block: Try this logic with transaction that does a transfer from an account with insufficient funds.

forester/tests/empty_address_tree_test.rs
```diff
@@ -59,22 +65,25 @@ async fn empty_address_tree_test() {
...
-        .await
-        .unwrap();
-    assert_eq!(get_state_queue_length(&config), 0);
+    let _ = empty_address_queue(
+        &mut env.rpc,
+        &mut env.indexer,
+        &env_accounts.forester,
+        &config,
+    )
+    .await;
```
ananas-block, line 80: one more unused result

test-utils/src/rpc/solana_rpc.rs
sergeytimoshin: Gotcha, thanks. I wrote a note for myself, let's fix this on another PR.

test-utils/src/spl.rs
ananas-block: agreed just saying

ananas-block (approved):

PR #812, opened 2024-06-09
feat: rust-sdk with account macro

SwenSchaeferjohann:
**newest changes**
* switched to buildjet for now as a quick fix to solving the 'running out of disk space' issue.
* most LOC changes come from (1) auto-generated idls and (2) removing dead code in the token_escrow program.
* the system-program and token-program now define instruction stubs under the #[build-idl] feature flag. This will be handy as this causes anchor to include the respective types that otherwise anchor would skip. This is unique to the idl-build feature (we previously used idl-parse). we're using this feature here (1) to enable our account macro usage; idl-build generates the IDL via compilation and is therefore able to apply our account proc macro before building the idl (2). We will be forced to use this anyway once we upgrade to anchor 0.30.0 (all programs need to be built using `idl-build`).

added: unit tests to the 2 new macros

**changes:**

Programs:
* remove anchorSerialize on structs that use zero_copy
** This allows us to use anchor's idl-build feature
** This should not impact our usage of those structs

Macros
* LightTraits macro, which auto-derives the trait implementations for relevant accounts
* LightAccounts macro, which extends a given struct with light's baseline fields

Light-SDK
* wrappers and util functions that aren't used in the system program but that should be used by consumers
* exec_verify method which wraps cpi preparation and invocation of the light system program
* other helpers for common, basic logic such as creating an output PDA
the token_escrow programs uses these ^


examples/token-escrow/programs/token-escrow/src/escrow_with_compressed_pda/withdrawal.rs
```diff
@@ -140,45 +144,12 @@ fn cpi_compressed_pda_withdrawal<'info>(
...
-        cpi_context_account,
-    };
-    let signer_seeds: [&[&[u8]]; 1] = [&seeds[..]];
-    let mut cpi_ctx = CpiContext::new_with_signer(
-        ctx.accounts.light_system_program.to_account_info(),
-        cpi_accounts,
-        &signer_seeds,
-    );
-    cpi_ctx.remaining_accounts = ctx.remaining_accounts.to_vec();
+    exec_verify(ctx, &inputs_struct, &[&signer_seeds])?;
```
ananas-block, line 151: naming choice is not immediately obvious to me.
What about one of the following?
1. invoke_light_system_program
2. invoke_system_program
3. verify_compressed_transaction
4. execute_compressed_transaction
5. invoke_with_signer
6. invoke_from_program
7. invoke_program
8. invoke_cpi

5 is my favorite because the import path makes it clear that it is the light system program and it is the same lingo as the solana sdk.
6 and 7 are even more explicit though.
8 would be consist with system program naming.

ananas-block (commented):
Took a first sweep. The diff of the token escrow program is difficult to look at will need to look at it locally again.
The deletions in the token escrow program withdrawal were duplicate code, since tests still pass and are not edited?
Commented with naming suggestions on the exec_verify wrapper.
In general, thanks for cleaning up the token escrow example looks much better already!

examples/token-escrow/programs/token-escrow/src/escrow_with_compressed_pda/withdrawal.rs
SwenSchaeferjohann: of the above `invoke_cpi` fits well with the underlying cpi instruction, agreed. thing here is people need to call this every time they want to write to any state / to verify and apply their state transition. hence the intuition to provide a dedicated very high-level naming for this wrapper.
* `light_sdk::verify()` (pref, because shorter)
* `light_sdk::verify_state_transition()`
i like these bc they're most descriptive at a very high level

macros/light/src/lib.rs
```diff
@@ -45,3 +46,396 @@ pub fn heap_neutral(_: TokenStream, input: TokenStream) -> TokenStream {
...
+///     pub account_compression_authority: AccountInfo<'info>,
+///     pub account_compression_program: Program<'info, AccountCompression>,
+///     pub system_program: Program<'info, System>,
+/// }
+/// ```
+#[proc_macro_derive(
+    LightTraits,
+    attributes(self_program, fee_payer, authority, cpi_context)
+)]
+pub fn light_traits_derive(input: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
SwenSchaeferjohann: done

examples/token-escrow/programs/token-escrow/src/escrow_with_compressed_pda/withdrawal.rs
SwenSchaeferjohann: renamed to verify for now - let's revisit this

SwenSchaeferjohann:
let's merge this after #816 . this will make the diffs slightly less convoluted

ananas-block (approved):

PR #818, opened 2024-06-12
chore: pedro's pr test ci

SwenSchaeferjohann:


js/stateless.js/src/rpc.ts
```diff
@@ -153,39 +153,46 @@ async function getCompressedTokenAccountsByOwnerOrDelegate(
...
+
     const compressedAccount: CompressedAccountWithMerkleContext =
         createCompressedAccountWithMerkleContext(
             createMerkleContext(
-                account.tree!,
+                compressedAccountResult.merkleTree,
                 mockNullifierQueue,
-                account.hash.toArray(undefined, 32),
-                account.leafIndex,
+                compressedAccountResult.hash.toArray(undefined, 32),
```
ananas-block, line 171: I think we should be explicit here whether to convert to "be" or "le" instead of undefined.

ananas-block (approved):

PR #822, opened 2024-06-13
feat: Add option to limit ledger size for test-validator

sergeytimoshin:


cli/src/utils/initTestEnv.ts
```diff
@@ -184,7 +190,7 @@ export async function getSolanaArgs({
 
   const solanaArgs = [
     "--reset",
-    `--limit-ledger-size=${LIMIT_LEDGER_SIZE}`,
+    `--limit-ledger-size=${limitLedgerSize}`,
```
ananas-block, line 193: what happens in the undefined case?
``--limit-ledger-size=undefined``

sergeytimoshin: If we don't pass limit-ledger-size we use the default value https://github.com/Lightprotocol/light-protocol/pull/822/files#diff-22aab76027e7cb6d8dd790150df31b957fb793269cfa29dda5539d2d4af55c41R57

ananas-block (approved):

PR #827, opened 2024-06-13
chore: rename gnark-prover to light-prover and circuitlib-rs to light-prover-client

sergeytimoshin:


test-utils/src/indexer/test_indexer.rs
```diff
@@ -617,8 +609,7 @@ impl<const INDEXED_ARRAY_SIZE: usize, R: RpcConnection> TestIndexer<INDEXED_ARRA
         let mut retries = 5;
         while retries > 0 {
             if retries < 3 {
-                spawn_gnark_server(self.path.as_str(), true, self.proof_types.as_slice()).await;
-                tokio::time::sleep(Duration::from_secs(10)).await;
+                spawn_prover(true, self.proof_types.as_slice()).await;
```
ananas-block, line 612: Not relevant right now but I noticed that since we want to ship test-utils as a rust testing suite for programs using zk compression we will need to refactor this eventually.
We can probably implement a trait for this as well so that we can pass in the options mono repo (use spawn prover) local testing (use light cli to spawn prover server)

ananas-block (approved):
Nice thanks for the improvement!
Feel free to merge after rebase and green ci.

PR #830, opened 2024-06-15
feat: get_multiple_new_address_proofs photon interop test

SwenSchaeferjohann:
* compare the address proofs of photon and test-indexer
* adds get_multiple_new_address_proofs to Indexer trait
* returns full height proof
* split interop tests into 2 test files, `nullify` and `address`
* asserts for pedro to debug photon responses with

* make env.create_addresses accept optional addresses


ananas-block (approved):
Nice to see you extending the e2e test env!

PR #837, opened 2024-06-18
chore: sync & update package.json dependencies 

sergeytimoshin:
1. TypeScript dependencies updated via `npx syncpack`
2. Fixed  `WARN  This project is configured to use v9.2.0 of pnpm. Your current pnpm is v9.1.0` for js/compressed-token


prover.js/package.json
```diff
@@ -36,36 +36,34 @@
...
-    "chai-as-promised": "^7.1.1",
+    "@types/mocha": "^10.0.6",
+    "@types/node": "^20.12.8",
+    "@typescript-eslint/eslint-plugin": "^7.6.0",
+    "@vitest/browser": "^1.6.0",
     "circomlib": "^2.0.5",
-    "eslint": "^8.50.0",
-    "eslint-config-standard-with-typescript": "^39.1.0",
+    "eslint": "8.57.0",
+    "eslint-config-love": "^53.0.0",
```
SwenSchaeferjohann, line 56: are we using this anywhere? @sergeytimoshin

sergeytimoshin: It should be a replacement for `eslint-config-standard-with-typescript`, but we don't use it anywhere.
Good catch, thanks!

SwenSchaeferjohann (approved):

PR #854, opened 2024-06-22
chore: avoid compilation errors on linux-arm64

sergeytimoshin:
This change helps users avoid compilation errors related to conflicting dependency versions.


ananas-block (approved):
Thanks!

PR #855, opened 2024-06-22
feat: add configurable ports and gossip-host to test-validator

sergeytimoshin:


cli/src/commands/start-prover/index.ts
```diff
@@ -27,6 +27,7 @@ class StartProver extends Command {
     loader.start();
 
     await startProver(
+      3001,
```
ananas-block, line 30: shouldn't this use the config?

ananas-block (approved):
Nitpick comment, approved optimistically if you need it quickly.

cli/src/commands/start-prover/index.ts
sergeytimoshin: fixed, thanks!

SwenSchaeferjohann (approved):

PR #859, opened 2024-06-23
refactor: remove isNative from token data, test wrapped sol

ananas-block:
### Issue:
- wrapped sol was not tested

### Solution:
- we just compress the wrapped sol token, the derivative not the SOL being represented
- this way there is no need to modify any transfer logic

### Note:
- token data has changed -> breaking changes for photon
- interop tests are disabled
- cli tests are disabled

### Changes:
- renamed `createMint` -> `createTokenPool` (instruction in compressed token program)
- refactor token data hashing to work with Option<delegate> and frozen with hashed data
- add wrapped sol test

### DIFF
--- a/programs/compressed-token/src/process_transfer.rs
+++ b/programs/compressed-token/src/process_transfer.rs
@@ -5,8 +5,8 @@ use crate::{
     ErrorCode, TransferInstruction,
 };
 use account_compression::utils::constants::CPI_AUTHORITY_PDA_SEED;
-use anchor_lang::{prelude::*, AnchorDeserialize};
-use light_hasher::{DataHasher, Poseidon};
+use anchor_lang::{prelude::*, solana_program::program_error::ProgramError, AnchorDeserialize};
+use light_hasher::Poseidon;
 use light_heap::{bench_sbf_end, bench_sbf_start};
 use light_system_program::{
     invoke::processor::CompressedProof,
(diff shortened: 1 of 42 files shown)


programs/compressed-token/src/process_transfer.rs
```diff
@@ -517,7 +485,7 @@ pub fn get_input_compressed_accounts_with_merkle_context_and_check_signer<const
         }
         let compressed_account = CompressedAccount {
             owner: crate::ID,
-            lamports: input_token_data.is_native.unwrap_or_default(),
+            lamports: 0,
```
SwenSchaeferjohann, line 488: @ananas-block 

If I understand this correctly, if we set input account lamports always to 0, compressed token accounts with lamports are not possible. that would be very useful to have, though, because it allows users to define spam filters for token accounts. 
also any output token accounts that were to be created with lamports would lead to awkward situations when trying to spend them 

fix: allow passing an optional `native_amount` in `input_token_data.`

SwenSchaeferjohann (commented):
one comment wrt input account lamports, otherwise looks great

programs/compressed-token/src/process_transfer.rs
ananas-block: sure I can add that

vadorovsky (approved):

SwenSchaeferjohann (approved):

PR #908, opened 2024-06-28
feat: anchor 0.30.1

sergeytimoshin:
This is important: https://github.com/Lightprotocol/light-protocol/pull/908/files#diff-0232306f21e881bf09e02d5acaec1ffc806ab5994d156514dc5ebe208184ea57R26-R30
Otherwise anchor will regenerate program ids on build


js/compressed-token/src/idl/light_compressed_token.ts
acheroncrypto, line 1: Just FYI, it looks like this was copied from the IDL JSON file because it has snake_case names. It's expected you use camelCase in TypeScript, and the correct types (camelCased) should be in the `target/types` directory.

ananas-block: thanks!

PR #910, opened 2024-06-29
chore: add rollover to e2e tests

ananas-block:


programs/account-compression/src/processor/initialize_address_queue.rs
```diff
@@ -33,9 +33,19 @@ pub fn process_initialize_address_queue<'info>(
                 .map_err(ProgramError::from)?
                 + compute_rollover_fee(rollover_threshold, height, queue_rent)
                     .map_err(ProgramError::from)?;
-            if (rollover_fee * rollover_threshold * (2u64.pow(height))) / 100
-                <= queue_rent + merkle_tree_rent
+            if rollover_fee != queue_rent + merkle_tree_rent
+                && (rollover_fee * rollover_threshold * (2u64.pow(height))) / 100
+                    < queue_rent + merkle_tree_rent
```
sergeytimoshin, line 38: We can extract this check as a method and use it both here and https://github.com/Lightprotocol/light-protocol/pull/910/files#diff-7a19889faf47467ed0c17862cab0a5f41b7ecb8c856289ba378577dfd40290c6R37-R39

sergeytimoshin (approved):

programs/account-compression/src/processor/initialize_address_queue.rs
ananas-block: thanks it's unified

PR #926, opened 2024-07-01
chore: switch indexed array to vec

ananas-block:
Changes:
- replace constant sized array and remove const generic from IndexedArray struct
- this will enable us to use bigger indexed arrays in tests


ananas-block:
pls hold off until we have the audit commit

vadorovsky (approved):

PR #929, opened 2024-07-01
chore: review-cleanup

ananas-block:
Changes:
- removed unused errors
- nullify leaves rename `indices` -> `leaf_indices`
- register program replace address constraint with constraint with same functionality because anchor 0.30. idl generation doesn't like address constraints
- update_address_merkle_tree - add comments, move address indexed element down, replace sequence_number with `merkle_tree.sequence_number` after it has been incremented
- append_state refactor index_merkle_tree_account to start at zero and be incremented with every iteration
- invoke/processor.rs - rename `roots` to `input_compressed_account_roots`
- invoke/processor.rs - remove deduplication of sequence numbers, its not needed anymore since we enforce ordered output Merkle trees
- rename `fetch_roots` -> `fetch_input_compressed_account_roots`
- `hash_input_compressed_accounts` -> enforce input ordering because we allow Merkle trees to be inserted into hashed_pubkeys without checking whether these exist already
-  merkle tree append performance has regressed by 1 from 141 to 140 for unknown reasons (no changes in append_leaves in account compression program)

### DIFF
--- a/js/stateless.js/src/idls/light_system_program.ts
+++ b/js/stateless.js/src/idls/light_system_program.ts
@@ -42,6 +42,9 @@ export type LightSystemProgram = {
                     name: 'feePayer';
                     isMut: true;
                     isSigner: true;
+                    docs: [
+                        'Fee payer needs to be mutable to pay rollover and protocol fees.',
+                    ];
                 },
                 {
                     name: 'authority';
@@ -62,23 +65,36 @@ export type LightSystemProgram = {
                     name: 'accountCompressionAuthority';
                     isMut: false;
                     isSigner: false;
+                    docs: [
+                        'This pda is used to invoke the account compression program.',
+                    ];
                 },
                 {
                     name: 'accountCompressionProgram';
                     isMut: false;
                     isSigner: false;
+                    docs: ['Merkle trees.'];
                 },
                 {
                     name: 'solPoolPda';
                     isMut: true;
                     isSigner: false;
                     isOptional: true;
+                    docs: [
+                        'Sol pool pda is used to store compressed sol.',
+                        "It's only required when compressing or decompressing sol.",
+                    ];
                 },
                 {
                     name: 'decompressionRecipient';
                     isMut: true;
                     isSigner: false;
                     isOptional: true;
+                    docs: [
+                        'Only needs to be provided for decompression as a recipient for the',
+                        'decompressed sol.',
+                        'Compressed sol originate from authority.',
+                    ];
                 },
                 {
                     name: 'systemProgram';
@@ -100,6 +116,9 @@ export type LightSystemProgram = {
                     name: 'feePayer';
                     isMut: true;
                     isSigner: true;
+                    docs: [
+                        'Fee payer needs to be mutable to pay rollover and protocol fees.',
+                    ];
                 },
                 {
                     name: 'authority';
@@ -174,6 +193,9 @@ export type LightSystemProgram = {
                     name: 'feePayer';
                     isMut: true;
                     isSigner: true;
+                    docs: [
+                        'Fee payer needs to be mutable to pay rollover and protocol fees.',
+                    ];
                 },
                 {
                     name: 'authority';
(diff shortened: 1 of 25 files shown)


js/stateless.js/src/idls/light_system_program.ts
```diff
@@ -194,23 +216,36 @@ export type LightSystemProgram = {
...
                     isSigner: false;
+                    docs: ['Merkle trees.'];
                 },
                 {
                     name: 'solPoolPda';
                     isMut: true;
                     isSigner: false;
                     isOptional: true;
+                    docs: [
+                        'Sol pool pda is used to store compressed sol.',
```
SwenSchaeferjohann, line 235: doesn't actually store compressed sol but native sol

js/stateless.js/src/idls/light_system_program.ts
```diff
@@ -1044,23 +1081,36 @@ export const IDL: LightSystemProgram = {
...
                     isSigner: false,
+                    docs: [
+                        'This pda is used to invoke the account compression program.',
+                    ],
                 },
                 {
                     name: 'accountCompressionProgram',
                     isMut: false,
                     isSigner: false,
+                    docs: ['Merkle trees.'],
```
SwenSchaeferjohann, line 1092: unclear docs

js/stateless.js/src/idls/light_system_program.ts
ananas-block: Joa both not wrong, the native sol that has been compressed.
Makes sense to be explicit let me fix it in https://github.com/Lightprotocol/light-protocol/actions/runs/9754900751/job/26922536668?pr=931 since the tests are already green here.

programs/system/src/invoke/processor.rs
```diff
@@ -74,10 +74,10 @@ pub fn process<
     // Compress or decompress lamports ---------------------------------------------------
     bench_sbf_start!("cpda_process_compression");
     if inputs.compress_or_decompress_lamports.is_some() {
-        compress_or_decompress_lamports(&inputs, &ctx)?;
         if inputs.is_compress && ctx.accounts.get_decompression_recipient().is_some() {
             return err!(SystemProgramError::DecompressionRecipienDefined);
```
SwenSchaeferjohann, line 78: typo
```suggestion
            return err!(SystemProgramError::DecompressionRecipientDefined);
```

programs/system/src/invoke/sol_compression.rs
```diff
@@ -31,10 +31,8 @@ pub fn compress_or_decompress_lamports<
 ) -> Result<()> {
     if inputs.is_compress {
         compress_lamports(inputs, ctx)
-    } else if inputs.compress_or_decompress_lamports.is_some() {
-        decompress_lamports(inputs, ctx)
     } else {
-        Ok(())
+        decompress_lamports(inputs, ctx)
     }
 }
```
SwenSchaeferjohann, line 37: this is now by default always going to execute `decompress_lamports()`,no? 
`is_compress` is not optional, so the default usage is is_compress = false, compress_or_decompress_lamports = None.

programs/system/src/invoke/sol_compression.rs
```diff
@@ -61,9 +59,7 @@ pub fn decompress_lamports<
         None => return err!(SystemProgramError::DeCompressLamportsUndefinedForDecompressSol),
     };
 
-    transfer_lamports(&sol_pool_pda, &recipient, lamports)?;
-
-    Ok(())
```
SwenSchaeferjohann, line 66: does it make any difference to not emit Ok(()) here

programs/system/src/invoke/sol_compression.rs
ananas-block: The function yes but we only execute the function when `compress_or_decompress_lamports` is some.
Hence, this function will never be executed if we don't want to compress or decompress.

programs/system/src/invoke/sol_compression.rs
ananas-block: It's cleaner not to

js/stateless.js/src/idls/light_system_program.ts
ananas-block: This is an error in the idl the full comment reads:
```
/// CHECK: Account compression program is used to update state and address
    /// Merkle trees.
    ```

SwenSchaeferjohann (approved):
approving, required changes will be addressed in #931

PR #940, opened 2024-07-02
chore: clean up forester tests

sergeytimoshin:


forester/tests/empty_address_tree_test.rs
```diff
@@ -111,23 +55,74 @@ async fn empty_address_tree_test() {
...
 
+fn setup_logger() {
+    let _ = env_logger::Builder::from_env(
+        env_logger::Env::default().default_filter_or(LevelFilter::Info.to_string()),
+    )
+    .is_test(true)
+    .try_init();
+}
+
+async fn spawn_test_validator() {
```
ananas-block, line 66: why do we have the same functions in multiple test files?

ananas-block: First thought the `keypair_action_config` getters are duplicates too but they are all different right?
can you move `spawn_test_validator` for example into a test-utils file?

ananas-block (approved):

PR #957, opened 2024-07-04
chore: add hash set bench test

ananas-block:


vadorovsky (commented):
What's the difference between this and `cargo xtask bench`?

I didn't dig into differences yet, but I prefer the `xtask` command - feels more intuitive than manually running an ignored test.

PR #981, opened 2024-07-10
chore: input compressed accounts batch compat

ananas-block:
Change:
- add `QueueIndex` placeholder in MerkleContext to avoid breaking changes in the api when we switch to batched Merkle tree updates.


vadorovsky (approved):
Amazing, thanks!

PR #998, opened 2024-07-16
feat: forester: multiple trees + rollover support

sergeytimoshin:


forester/src/config.rs
```diff
@@ -1,38 +1,44 @@
...
+            address_tree_data: self.address_tree_data.clone(),
+            last_sync_data: self.last_sync_data.clone(),
         }
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct TreeSyncMetadata {
+    pub last_synced_at: DateTime<Utc>,
+    pub last_synced_root: [u8; 32],
```
ananas-block, line 43: what's the conceptual idea behind this?
how often do you sync trees?

forester/src/main.rs
```diff
@@ -103,48 +107,120 @@ async fn main() {
...
+) {
     let indexer_rpc = init_rpc(config.clone(), false).await;
     let indexer = Arc::new(tokio::sync::Mutex::new(PhotonIndexer::new(
         config.external_services.indexer_url.to_string(),
         indexer_rpc,
     )));
 
-    nullify_state(config.clone(), rpc_pool, indexer).await;
+    nullify_addresses(config.clone(), rpc_pool, indexer, tree_data, rollover_state).await;
 }
```
ananas-block, line 211: Are these still used?
It looks like `run_subscribe` covers the same functionality unified right?

forester/src/rollover/operations.rs
ananas-block, line 1: really would like to move code like this which is duplicate in forester and test-utils into a forester-utils crate which forester and test-utils depend on.

sergeytimoshin: I started to extract duplicated code from forester & test-utils, and I think it would be better to make it in different PR, to many changes there.

forester/src/config.rs
sergeytimoshin: Removed it for now. 
I sync the trees on every start, but that's obviously not enough, and syncing on every rollover is not enough either. 
I see 2 options:
1. Separate thread that would subscribe to account_compression account changes and do a sync on each change.
2. Just a cron job, every N minutes.

sergeytimoshin: Wdyt?

ananas-block: The forester epochs and slots will change this a bit. (Foresters are only eligible to forest during their own slots.)
It’s probably enough to sync the tree leading up and during the slot.

ananas-block (approved):

PR #1002, opened 2024-07-17
refactor: get_anchor_account returns Result<Option<T>, RpcError>> instead of unwrapped T

sergeytimoshin:


test-utils/src/rpc/rpc_connection.rs
```diff
@@ -25,12 +25,25 @@ pub trait RpcConnection: Clone + Send + Sync + Debug + 'static {
...
-        authority: &Pubkey,
-        signers: &[&Keypair],
-    ) -> impl std::future::Future<Output = Result<Signature, RpcError>> + Send;
+    fn create_and_send_transaction<'a>(
+        &'a mut self,
+        instruction: &'a [Instruction],
+        payer: &'a Pubkey,
+        signers: &'a [&'a Keypair],
+    ) -> impl std::future::Future<Output = Result<Signature, RpcError>> + Send + 'a {
+        async move {
```
ananas-block, line 34: whats the difference between `async move` and `async fn`?

ananas-block (approved):

test-utils/src/rpc/rpc_connection.rs
sergeytimoshin: `async move` is just an anonymous clojure inside the fn body. We can't use `async fn`, clippy doesn't like it in traits, so we use `impl Future<..>` instead.

PR #1031, opened 2024-07-25
feat: forester: add PHOTON_API_KEY .env var support

sergeytimoshin:


photon-api/src/apis/default_api.rs
```diff
@@ -956,17 +991,23 @@ pub async fn get_multiple_compressed_account_proofs_post(
...
-    let local_var_uri_str = format!(
+    let mut local_var_uri_str = format!(
         "{}/getMultipleCompressedAccountProofs",
         local_var_configuration.base_path
     );
+
+    if let Some(ref api_key) = local_var_configuration.api_key {
+        local_var_uri_str = format!("{}?api_key={}", local_var_uri_str, api_key.key);
+    }
+
```
ananas-block, line 1002: can we use `append_api_key` here as well?

ananas-block (commented):
Thanks, looks good in general just one comment.

photon-api/src/apis/default_api.rs
sergeytimoshin: sure, thanks!

ananas-block (approved):

PR #1033, opened 2024-07-25
chore: Formal Specification for the Merkle Tree circuits

kustosz:
This includes:
1. Updates to the circuit-to-lean extraction command so that it actually exports the circuit
2. A library of proofs and theorems, with the main correctness results stated in Main.lean
3. Updates to the CI pipeline, to run the extraction & Lean verification on each CI run
4. Some optimizations for the circuits – removing unnecessary constraints without changing the circuit semantics


light-prover/prover/non_inclusion_circuit.go
```diff
@@ -39,10 +39,7 @@ func (circuit *NonInclusionCircuit) Define(api frontend.API) error {
 		NumberOfCompressedAccounts: circuit.NumberOfCompressedAccounts,
 		Depth:                      circuit.Depth,
 	}
-	roots := abstractor.Call1(api, proof)
-	for i := 0; i < int(circuit.NumberOfCompressedAccounts); i++ {
-		api.AssertIsEqual(roots[i], circuit.Roots[i])
-	}
+	abstractor.Call1(api, proof)
```
ananas-block, line 42: Nice, I like that it's simpler now!

light-prover/prover/extractor.go
```diff
@@ -13,20 +13,39 @@ func ExtractLean(treeDepth uint32, numberOfCompressedAccounts uint32) (string, e
 	roots := make([]frontend.Variable, numberOfCompressedAccounts)
 	leaves := make([]frontend.Variable, numberOfCompressedAccounts)
 	inPathIndices := make([]frontend.Variable, numberOfCompressedAccounts)
```
ananas-block, line 15: shouldn't we have separate `inPathIndices`for `inclusion` and `non-inclusion`?
(right now it seems like the same `inPathIndices` are reused for both `inclusion` (line 29)  and `non-inclusion` (line 41).)

ananas-block (commented):
Thank you!
Awesome that you integrated it into CI already!
I like the simplification you did in the non-inclusion circuit.
Looks good in general, just a couple questions, readme how to run tests locally would be great and one variable is used twice in ExtractLean which looks suspicious.

.github/workflows/prover-test.yml
```diff
@@ -53,3 +64,14 @@ jobs:
         run: |
           cd light-prover
           go test -timeout 30m
+
+      - name: Extract circuit to Lean
+        run: |
+          cd light-prover
+          ./light-prover extract-circuit --output formal-verification/FormalVerification/Circuit.lean --tree-depth=20 --compressed-accounts=10
```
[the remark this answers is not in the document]
kustosz: This command extracts the definition from golang into the `circuit.lean` file – that is it makes sure we actually work based on your implementation and not our idea of it. `Main.lean` imports that file and proves the theorems. It will stop compiling if `Circuit.lean` changes (which will happen if the circuit definition in golang changes).

Depth and number of inputs have no impact on any of this – I've changed it to 26/8 for you and the only changes I've had to make was find-and-replace the autogenerated circuit names.

ananas-block (approved):

PR #1050, opened 2024-08-04
fix: mix delegated and owned compressed token accounts 

ananas-block:
Issue:
- consider an escrow program with an compressed escrow token account owned by a pda
- when compressing into this escrow account we have token accounts of two different owners which both need to sign (pda, depositing user)
- The straightforward way is to just transfer the deposit amount to the pda. However, this way every deposit creates an additional escrow account -> the escrowed tokens are fragmented and need to be merged.
- We could merge escrow accounts in the deposit transaction in a second cpi. The implementation is difficult since we need to dynamically get the leaf index of the newly inserted compressed account onchain, then spend the newly created compressed account and verify two zkps in this process. This would only work if there is no activity on the Merkle tree otherwise the second zkp will fail.
- A feasible workaround is approve the funds to escrow and transfer the approved funds to the escrow token account (2 inputs, escrow, approved token account, 1 output escrow)

Changes:
- enable two owners for input compressed accounts

### DIFF
--- a/test-programs/compressed-token-test/tests/test.rs
+++ b/test-programs/compressed-token-test/tests/test.rs
@@ -1164,10 +1164,6 @@ async fn test_delegation_mixed() {
             test_indexer.get_compressed_token_accounts_by_owner(&delegate.pubkey());
 
         input_compressed_accounts.extend_from_slice(&delegate_input_compressed_accounts);
-        // let lamports_output_amount = input_compressed_accounts
-        //     .iter()
-        //     .map(|x| x.compressed_account.compressed_account.lamports)
-        //     .sum::<u64>();
         let input_amount = input_compressed_accounts
             .iter()
             .map(|x| x.token_data.amount)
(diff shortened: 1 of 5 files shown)


test-programs/compressed-token-test/tests/test.rs
```diff
@@ -990,9 +992,203 @@ async fn test_delegation(
...
+            .cloned()
+            .collect::<Vec<TokenDataWithContext>>();
+        let delegate_input_compressed_accounts =
+            test_indexer.get_compressed_token_accounts_by_owner(&delegate.pubkey());
+
+        input_compressed_accounts.extend_from_slice(&delegate_input_compressed_accounts);
+        // let lamports_output_amount = input_compressed_accounts
+        //     .iter()
+        //     .map(|x| x.compressed_account.compressed_account.lamports)
+        //     .sum::<u64>();
```
SwenSchaeferjohann, line 1170: comment?

SwenSchaeferjohann (approved):

test-programs/compressed-token-test/tests/test.rs
ananas-block, line 1170:
```suggestion
```

PR #1068, opened 2024-08-08
feat: optimize forester with concurrent proof fetching and tx processing

sergeytimoshin:
- Implement parallel fetching of proofs from indexer
- Add concurrent processing of nullification transactions
- Introduce chunked work item processing with backpressure
- Add TPS measurement for performance monitoring


ananas-block (approved):
Looks good to me, and I don't want to hold you back.

PR #1072, opened 2024-08-10
chore: RPC patch

SwenSchaeferjohann:
* bump versions to latest NPM releases
* changes the JS RPC return type for endpoints with cursors. Wraps inside WithCursor<{items, cursor}>, so values need to be accessed like e.g. compressedAccounts.items[0]

### DIFF
--- a/js/compressed-token/package.json
+++ b/js/compressed-token/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@lightprotocol/compressed-token",
-    "version": "0.4.1",
+    "version": "0.7.0",
     "description": "JS client to interact with the compressed-token program",
     "sideEffects": false,
     "main": "dist/cjs/node/index.cjs",
--- a/js/compressed-token/package.json
+++ b/js/compressed-token/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@lightprotocol/compressed-token",
-    "version": "0.8.0",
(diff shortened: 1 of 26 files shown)


js/compressed-token/package.json
```diff
@@ -1,28 +1,23 @@
 {
     "name": "@lightprotocol/compressed-token",
-    "version": "0.4.1",
+    "version": "0.8.0",
```
ananas-block, line 3: current release is 0.10.1

js/stateless.js/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@lightprotocol/stateless.js",
-    "version": "0.5.1",
+    "version": "0.9.0",
```
ananas-block, line 3: current version is 0.10.1

ananas-block (approved):
Apart from the two incorrect versions lgtm.

js/stateless.js/package.json
ananas-block:
```suggestion
    "version": "0.10.1",
```

js/compressed-token/package.json
ananas-block:
```suggestion
    "version": "0.10.1",
```

vadorovsky (approved):

PR #1089, opened 2024-08-14
feat: forester: do not retry on not eligible slots

sergeytimoshin:
Introduced a new error variant `NotEligible` in `ForesterError`. 
Updated the eligibility check to return this error and handle it appropriately during the transaction processing loop.

### DIFF
--- a/forester/src/epoch_manager.rs
+++ b/forester/src/epoch_manager.rs
@@ -720,7 +720,7 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
         &self,
         registration_info: &ForesterEpochInfo,
         tree_account: &TreeAccounts,
-    ) -> Result<bool> {
+    ) -> Result<()> {
         let mut rpc = self.rpc_pool.get_connection().await?;
         let current_slot = rpc.get_slot().await?;
         let forester_epoch_pda = rpc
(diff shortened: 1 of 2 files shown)


forester/src/epoch_manager.rs
```diff
@@ -873,48 +877,55 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
...
+                        }
+                        Err(e) => {
+                            if retries >= self.config.max_retries {
+                                error!(
+                                    "Max retries reached for work item {:?}. Error: {:?}",
+                                    work_item.queue_item_data.hash, e
+                                );
+                                return Err(e);
+                            }
+                            let delay = BASE_RETRY_DELAY * 2u32.pow(retries as u32);
```
SwenSchaeferjohann, line 911: theoretic nit: could use saturating_pow to prevent overflows. anyway don't think it's important because retrys are not gonna overflow in practice

SwenSchaeferjohann (approved):
lgtm!

forester/src/epoch_manager.rs
```diff
@@ -873,48 +877,55 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
...
+                            sleep(delay + Duration::from_millis(jitter)).await;
+                            retries += 1;
+                            warn!(
+                                "Retrying work item {:?}. Attempt {}/{}",
+                                work_item.queue_item_data.hash, retries, self.config.max_retries
+                            );
+                        }
+                    }
+                }
+                Err(ForesterError::NotEligible) => {
```
ananas-block, line 922: Nice, are there other.
Are there other errors that we need to handle?
For example, edge cases changing epochs before start or after it has ended (might be covered haven't checked).

sergeytimoshin: Yes, there are still many places in forester where I use Forester::Custom(String) generic errors, I should definitely use less generic errors for the specific cases. It's on my list.

forester/src/epoch_manager.rs
sergeytimoshin: Fixed.

PR #1092, opened 2024-08-15
chore: program anchor 0.30.1 compat

ananas-block:


programs/compressed-token/src/instructions/freeze.rs
```diff
@@ -1,32 +1,36 @@
...
+
+use crate::program::LightCompressedToken;
 
 #[derive(Accounts)]
 pub struct FreezeInstruction<'info> {
     /// UNCHECKED: only pays fees.
     #[account(mut)]
     pub fee_payer: Signer<'info>,
-    #[account(address= mint.freeze_authority.unwrap()
+    #[account(constraint= authority.key() == mint.freeze_authority.unwrap()
```
SwenSchaeferjohann, line 16: what is the difference between these?

macros/light/src/accounts.rs
```diff
@@ -10,12 +10,12 @@ pub(crate) fn process_light_accounts(input: DeriveInput) -> Result<TokenStream>
...
-                    "Program<'info, ::light_system_program::program::LightSystemProgram>",
+                    "Program<'info, ::LightSystemProgram>",
                 ),
                 ("system_program", "Program<'info, System>"),
                 (
                     "account_compression_program",
-                    "Program<'info, ::account_compression::program::AccountCompression>",
+                    "Program<'info, ::AccountCompression>",
                 ),
             ];
```
SwenSchaeferjohann, line 20: i believe this change causes issues when used by programs that don't import account_compression themselves. reverting the change should make token_escrow compile again IIRC

programs/compressed-token/src/instructions/freeze.rs
ananas-block: address is not supported in anchor 0.30.1

SwenSchaeferjohann (approved):
lgmt. feel free to merge when green

PR #1095, opened 2024-08-18
feat: forester alerts + monitoring

sergeytimoshin:
Added:

• Tracing-based telemetry
• Prometheus metrics sender
• Prometheus + AlertManager + PushGateway + Grafana docker-compose configuration

### DIFF
--- a/forester/docker-compose.yml
+++ b/forester/docker-compose.yml
@@ -1,5 +1,3 @@
-name: photondb
-
 services:
   db:
     image: postgres:16-alpine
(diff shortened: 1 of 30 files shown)


forester/docker-compose.yml
```diff
@@ -22,7 +20,71 @@ services:
       - PGADMIN_DEFAULT_PASSWORD=photon
```
ananas-block, line 20: should we make this an env variable?
(Same question for postgres login above.)

forester/docker-compose.yml
```diff
@@ -22,7 +20,71 @@ services:
...
+      - ./alert_rules.yml:/etc/prometheus/alert_rules.yml
+    networks:
+      - forester-net
+
+  grafana:
+    image: grafana/grafana
+    ports:
+      - "3000:3000"
+    environment:
+      - GF_SECURITY_ADMIN_PASSWORD=admin
```
ananas-block, line 78: env variable necessary?

forester/src/epoch_manager.rs
```diff
@@ -541,11 +527,24 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
                     epoch_pda.epoch,
                 )
                 .await?;
+
+                // Prometheus metrics
+                let chunk_duration = start_time.elapsed();
+                queue_metric_update(epoch_info.epoch, num_tx_sent, chunk_duration).await;
+
+                // TODO: consider do we really need WorkReport
+                self.increment_processed_items_count(epoch_info.epoch, num_tx_sent)
```
ananas-block, line 536: I see that we could use it to push things to graphana but in once we have multiple foresters by different parties we should get the performed work from the `ForesterEpochPda`s and the total performed work from the `EpochPda` .

test-utils/src/rpc/solana_rpc.rs
```diff
@@ -345,6 +361,7 @@ impl RpcConnection for SolanaRpcConnection {
     }
 
     async fn get_slot(&mut self) -> Result<u64, RpcError> {
+        println!("Calling get_slot");
```
ananas-block, line 364: shouldn't we call `info!` here?

forester/docker-compose.yml
sergeytimoshin: yes we can, and probably should, it's just a default test value, it implies that the end user should replace it with something more secure

ananas-block (approved):

PR #1132, opened 2024-08-31
feat: add test-validator --stop flag

SwenSchaeferjohann:
make it easy for users to stop the test-validator and related background processes

### DIFF
--- a/cli/src/commands/test-validator/index.ts
+++ b/cli/src/commands/test-validator/index.ts
@@ -1,5 +1,5 @@
 import { Command, Flags } from "@oclif/core";
-import { initTestEnv } from "../../utils/initTestEnv";
+import { initTestEnv, stopTestEnv } from "../../utils/initTestEnv";
 import { CustomLoader } from "../../utils/index";
 
 class SetupCommand extends Command {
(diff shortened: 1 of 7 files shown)


cli/src/commands/test-validator/index.ts
```diff
@@ -80,29 +80,43 @@ class SetupCommand extends Command {
       required: false,
       default: "127.0.0.1",
     }),
+    stop: Flags.boolean({
+      description: "Stops the test validator and dependent processes.",
+      required: false,
+      default: false,
```
sergeytimoshin, line 86: It's not critical, but I think it makes sense to add the `exclusive: [<all other available flags>]` parameter or even make this a standalone command.

cli/src/utils/initTestEnv.ts
```diff
@@ -17,9 +17,45 @@ import {
...
+      condition: true,
+      killFunction: killTestValidator,
+    },
+  ];
+
+  for (const process of processesToKill) {
+    if (process.condition) {
+      try {
+        if (process.killFunction) {
+          await process.killFunction();
```
sergeytimoshin, line 48: We can use array of promises here to stop all the processes at once.

cli/src/commands/test-validator/index.ts
SwenSchaeferjohann: user can pass eg `--stop --skip-prover` to keep those alive

SwenSchaeferjohann: gonna add to 'description' for clarity

sergeytimoshin (approved):

PR #1134, opened 2024-09-01
chore: added rollover logic to forester(once per slot) and test

ananas-block:
Issue:
- foresters only check once per active phase for every tree whether it can be rolledover, for long epochs with high traffic this is insufficient

Changes:
- check whether a tree can be rolledover once per light slot
- move rollover  into `process_queue`
- modify test to test a rollover

### DIFF
--- a/forester/src/rollover/operations.rs
+++ b/forester/src/rollover/operations.rs
@@ -52,7 +52,8 @@ pub async fn is_tree_ready_for_rollover<R: RpcConnection>(
                 .get_anchor_account::<StateMerkleTreeAccount>(&tree_pubkey)
                 .await?
                 .unwrap();
-            info!("Account: {:?}", account);
+            // let account_info = rpc.get_account(tree_pubkey).await?.unwrap();
+
             let is_already_rolled_over =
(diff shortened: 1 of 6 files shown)


forester/src/rollover/operations.rs
```diff
@@ -68,14 +69,24 @@ pub async fn is_tree_ready_for_rollover<R: RpcConnection>(
...
                 .unwrap();
-            info!("Account: {:?}", account);
+            let queue_account = rpc
+                .get_anchor_account::<QueueAccount>(&account.metadata.associated_queue)
+                .await?
+                .unwrap();
+            // let account_info = rpc
+            //     .get_account(account.metadata.associated_queue)
+            //     .await?
+            //     .unwrap();
```
sergeytimoshin, line 89: can we remove it?

sergeytimoshin (approved):

forester/src/rollover/operations.rs
ananas-block: I left it on purpose so that we know that we need to take the queue account info not the Merkle tree.
This is different for state and address Merkle trees because for address Merkle trees users only interact with the queue the tree is only updated by the forester.

PR #1143, opened 2024-09-02
refactor: forester: switch logging from `log` to `tracing`

sergeytimoshin:
* Replaced the `log` crate with `tracing` for improved structured logging. 
* Adjusted dependencies in `Cargo.toml` and removed `log` crate.

### DIFF
--- a/forester/src/lib.rs
+++ b/forester/src/lib.rs
@@ -28,12 +28,12 @@ pub use config::{ForesterConfig, ForesterEpochInfo};
 use forester_utils::forester_epoch::{TreeAccounts, TreeType};
 use forester_utils::indexer::Indexer;
 use forester_utils::rpc::{RpcConnection, SolanaRpcConnection};
-use log::info;
 pub use settings::init_config;
 use solana_sdk::commitment_config::CommitmentConfig;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{mpsc, oneshot, Mutex};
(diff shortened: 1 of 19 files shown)


forester/src/lib.rs
```diff
@@ -55,7 +55,7 @@ pub async fn run_queue_info(
         QUEUE_LENGTH
             .with_label_values(&[&*queue_type.to_string(), &tree_data.merkle_tree.to_string()])
             .set(queue_length as i64);
-        info!(
+        println!(
```
ananas-block, line 58: `println!` instead of `debug` on purpose?

forester/src/telemetry.rs
```diff
@@ -13,16 +13,35 @@ pub fn setup_telemetry() {
...
 
+        // tracing_subscriber::registry()
+        //     .with(env_filter)
+        //     .with(
+        //         fmt::Layer::new()
+        //             .with_writer(std::io::stdout)
+        //             .with_ansi(true),
+        //     )
+        //     .with(fmt::Layer::new().with_writer(non_blocking).json())
+        //     .init();
```
ananas-block, line 43: can we remove?

forester/src/lib.rs
sergeytimoshin: Yes. It's not a debug printout, its output for the user.

ananas-block (approved):

PR #1160, opened 2024-09-06
refactor: forester: add retry logic for epoch registration

sergeytimoshin:
Consolidate repeated active phase work processing into a new function `process_epoch_work`. 
Introduce `register_for_epoch_with_retry` to handle registration retries with a specified maximum number of attempts and delay duration.

### DIFF
--- a/forester/src/epoch_manager.rs
+++ b/forester/src/epoch_manager.rs
@@ -222,48 +222,36 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
         let previous_phases =
             get_epoch_phases(&self.protocol_config, current_epoch.saturating_sub(1));
 
-        // Check if we're in the active phase of the previous epoch
+        // Process previous epoch if still in active phase
         if slot >= previous_phases.active.start && slot < previous_phases.active.end {
             info!(
-                "Currently in active phase of previous epoch {}. Processing remaining work.",
+                "Processing remaining work for previous epoch {}",
                 current_epoch - 1
             );
-
-            match self.recover_registration_info(current_epoch - 1).await {
-                Ok(previous_registration_info) => {
-                    self.perform_active_work(&previous_registration_info)
-                        .await?;
-                }
-                Err(e) => {
-                    warn!(
-                        "Failed to recover registration info for previous epoch {}: {:?}",
-                        current_epoch - 1,
-                        e
-                    );
-                }
-            }
+            self.process_epoch_work(current_epoch - 1).await?;
         }
 
-        // Check if we're in the active phase of the current epoch
+        // Process current epoch
         if slot >= current_phases.active.start && slot < current_phases.active.end {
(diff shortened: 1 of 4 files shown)


ananas-block (changes requested):
Further changes are necessary to handle not only active work but subsequent epoch steps correctly for recovered epochs.

forester/src/epoch_manager.rs
```diff
@@ -272,24 +257,61 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
...
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .is_err()
+        {
+            // Another task is already processing this epoch
+            debug!("Epoch {} is already being processed, skipping", epoch);
+            return Ok(());
+        }
+
+        // Attempt to recover registration info
+        let mut registration_info = match self.recover_registration_info(epoch).await {
```
ananas-block, line 275: what does this do if there is no registration that can be recovered?

forester/src/epoch_manager.rs
```diff
@@ -299,6 +321,52 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
...
+        let mut rpc = self.rpc_pool.get_connection().await?;
+        let slot = rpc.get_slot().await?;
+        let phases = get_epoch_phases(&self.protocol_config, epoch);
+
+        // Check if it's already too late to register
+        if slot >= phases.registration.end {
+            return Err(ForesterError::Custom(format!(
+                "Too late to register for epoch {}. Current slot: {}, Registration end: {}",
+                epoch, slot, phases.registration.end
+            )));
```
ananas-block, line 341: do we really want to throw in this case?
An alternative behavior could be to wait for the next registration period.
Or just return if we have logic to wait for the next epoch in a different place.

forester/src/epoch_manager.rs
sergeytimoshin: If there is no registration we'll try to register here https://github.com/Lightprotocol/light-protocol/pull/1160/files/7213266f76690b8931a73e01055af9123b07b22a#diff-1e165bd3189acab768bdb1cd6cf4ec33528400511ec68498004b73a25d0646c3R280-R281

forester/src/epoch_manager.rs
sergeytimoshin: I think this behaviour is semantically correct because it's part of the process_epoch flow: this case is really an error, it shouldn't happen in a normal situation, and we log it as an error in our logfiles: https://github.com/Lightprotocol/light-protocol/pull/1160/files/7213266f76690b8931a73e01055af9123b07b22a#diff-1e165bd3189acab768bdb1cd6cf4ec33528400511ec68498004b73a25d0646c3R147

ananas-block (approved):

PR #1166, opened 2024-09-08
feat: add retry logic to RPC methods

sergeytimoshin:
Added the 'async-trait' crate to provide async trait methods in RPC interfaces. 
Implemented async versions of existing synchronous methods and added retry logic for better error handling in 'SolanaRpcConnection'.


forester-utils/src/rpc/solana_rpc.rs
```diff
@@ -272,116 +337,108 @@ impl RpcConnection for SolanaRpcConnection {
...
+            self.client
+                .send_transaction_with_config(
+                    transaction,
+                    RpcSendTransactionConfig {
+                        skip_preflight: true,
+                        max_retries: Some(self.retry_config.max_retries as usize),
+                        ..Default::default()
+                    },
+                )
+                .map_err(RpcError::from)
```
ananas-block, line 440: we use this logic in `send_transaction` in forester which already has retry logic, is that retry logic now redundant?

sergeytimoshin: Yes. We need retry logic for each rpc method (otherwise we can easily throw out of from process_queue), so I decided to consolidate it in one place.

ananas-block (approved):

PR #1170, opened 2024-09-08
refactor: forester: add trees in wait_for_active_phase for recovery + don't retry to send report tx if already sent

sergeytimoshin:
* Do not check phases for wait_for_active_phase
* Handle report_work transaction errors gracefully in Epoch Manager

### DIFF
--- a/forester/src/epoch_manager.rs
+++ b/forester/src/epoch_manager.rs
@@ -797,6 +797,25 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
         info!("Reporting work");
         let mut rpc = self.rpc_pool.get_connection().await?;
 
+        let forester_epoch_pda_pubkey = get_forester_epoch_pda_from_authority(
+            &self.config.payer_keypair.pubkey(),
+            epoch_info.epoch.epoch,
+        )
+        .0;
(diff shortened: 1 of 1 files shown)


forester/src/epoch_manager.rs
```diff
@@ -799,12 +801,43 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
             &self.config.payer_keypair.pubkey(),
             epoch_info.epoch.epoch,
         );
-        rpc.create_and_send_transaction(
```
ananas-block, line 802: would it make sense to do an additional sanity check whether work is already reported by fetching `forester_epoch_pda` and checking its `has_reported_work` field?

forester/src/epoch_manager.rs
```diff
@@ -795,16 +797,66 @@ impl<R: RpcConnection, I: Indexer<R>> EpochManager<R, I> {
         info!("Reporting work");
         let mut rpc = self.rpc_pool.get_connection().await?;
 
+        let forester_epoch_pda_pubkey = get_forester_epoch_pda_from_authority(
```
ananas-block, line 800: we can get this value from `epoch_info.forester_epoch_pda`

ananas-block (approved):

PR #1171, opened 2024-09-08
fix: typescript sdk: use cursor when accumulating accounts for transfer

sergeytimoshin:
If we have > than 1000 accounts with 0 balance, we need to fetch accounts with cursor.

### DIFF
--- a/js/stateless.js/src/actions/transfer.ts
+++ b/js/stateless.js/src/actions/transfer.ts
@@ -47,17 +47,17 @@ export async function transfer(
     /// TODO: allow multiple
     merkleTree?: PublicKey,
     confirmOptions?: ConfirmOptions,
-    config?: Omit<GetCompressedAccountsByOwnerConfig, 'cursor' | 'limit'>,
 ): Promise<TransactionSignature> {
     let accumulatedLamports = bn(0);
     const compressedAccounts: CompressedAccountWithMerkleContext[] = [];
-    let cursor: string | null = null;
+    let cursor: string | undefined;
     const batchSize = 1000; // Maximum allowed by the API
     lamports = bn(lamports);
 
     while (accumulatedLamports.lt(lamports)) {
         const batchConfig: GetCompressedAccountsByOwnerConfig = {
-            ...config,
+            filters: undefined,
+            dataSlice: undefined,
             cursor,
             limit: new BN(batchSize),
         };
@@ -73,7 +73,7 @@ export async function transfer(
             }
         }
 
-        cursor = batch.cursor;
+        cursor = batch.cursor ?? undefined;
         if (batch.items.length < batchSize || accumulatedLamports.gte(lamports))
             break;
     }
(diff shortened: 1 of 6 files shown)


ananas-block (approved):

js/stateless.js/src/rpc.ts
```diff
@@ -664,7 +664,7 @@ export class Rpc extends Connection implements CompressionApiInterface {
      */
     async getCompressedAccountsByOwner(
         owner: PublicKey,
-        config?: GetCompressedAccountsByOwnerConfig,
+        config?: GetCompressedAccountsByOwnerConfig | undefined,
```
SwenSchaeferjohann, line 667: can remove `undefined`, `?` already takes care of it

js/stateless.js/src/rpc-interface.ts
```diff
@@ -504,7 +504,9 @@ export interface CompressionApiInterface {
 
     getCompressedAccountsByOwner(
         owner: PublicKey,
-        config?: GetCompressedAccountsByOwnerConfig,
+        config?:
+            | Omit<GetCompressedAccountsByOwnerConfig, 'cursor' | 'limit'>
+            | undefined,
```
SwenSchaeferjohann, line 509: can remove `undefined`, `?` already takes care of it

why exclude `cursor` and `limit` from the param?

js/stateless.js/src/rpc-interface.ts
```diff
@@ -40,7 +40,7 @@ export interface LatestNonVotingSignatures {
 export interface GetCompressedAccountsByOwnerConfig {
     filters?: GetCompressedAccountsFilter[];
     dataSlice?: DataSlice;
-    cursor?: string;
+    cursor: string | null;
```
SwenSchaeferjohann, line 43: why replace `undefined` with `null`? this makes the interface inconsistent

js/stateless.js/src/test-helpers/test-rpc/test-rpc.ts
```diff
@@ -335,9 +335,11 @@ export class TestRpc extends Connection implements CompressionApiInterface {
...
     async getCompressedAccountsByOwner(
         owner: PublicKey,
-        _config?: GetCompressedAccountsByOwnerConfig,
+        config?:
+            | Omit<GetCompressedAccountsByOwnerConfig, 'cursor' | 'limit'>
+            | undefined,
     ): Promise<WithCursor<CompressedAccountWithMerkleContext[]>> {
-        if (_config) {
+        if (config) {
             throw new Error(
```
SwenSchaeferjohann, line 343: in which scenario would test-rpc reach >1000 events?

js/stateless.js/src/actions/transfer.ts
```diff
@@ -40,14 +47,45 @@ export async function transfer(
     /// TODO: allow multiple
     merkleTree?: PublicKey,
     confirmOptions?: ConfirmOptions,
+    config?: Omit<GetCompressedAccountsByOwnerConfig, 'cursor' | 'limit'>,
```
SwenSchaeferjohann, line 50: what's the reasoning for passing filter params along to the `transfer` action?

SwenSchaeferjohann (changes requested):

SwenSchaeferjohann (approved):

PR #1194, opened 2024-09-09
chore: add deployment scripts

ananas-block:


vadorovsky (approved):
The content looks good, but the PR title looks unrelated? I see scripts for deployment instead and no readme changes

PR #1195, opened 2024-09-10
refactor: forester: safetys check to minimise non-eligibility + other safety checks

sergeytimoshin:
Improve timeout handling in epoch manager to prevent overflow errors. 
Introduce a checked multiplication for slot duration and update retry timeout accordingly. 
Additionally, modify transaction handling to break early based on calculated slot latency.

### DIFF
--- a/forester/src/send_transaction.rs
+++ b/forester/src/send_transaction.rs
@@ -25,7 +25,8 @@ use solana_sdk::{
     signature::{Keypair, Signer},
 };
 use std::sync::Arc;
-use std::{time::Duration, vec};
+use std::time::Duration;
+use std::vec;
 use tokio::join;
 use tokio::sync::Mutex;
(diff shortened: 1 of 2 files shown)


forester/src/send_transaction.rs
```diff
@@ -112,7 +120,19 @@ pub async fn send_batched_transactions<T: TransactionBuilder, R: RpcConnection>(
...
+                .checked_sub(start_time.elapsed())
+            {
+                Some(time) => time,
+                None => {
+                    debug!("Reached end of light slot");
+                    break;
+                }
+            };
+
+            if remaining_time < LATENCY {
```
ananas-block, line 135: we could also just make the timeout shorter

ananas-block (approved):

vadorovsky (approved):

PR #1200, opened 2024-09-11
refactor: forester: queue fetching logic to support configurable indices

sergeytimoshin:
Enhanced fetch_queue_item_data to accept start index, length, and queue length parameters.


forester/forester.toml
```diff
@@ -7,4 +7,8 @@ INDEXER_BATCH_SIZE=50
 INDEXER_MAX_CONCURRENT_BATCHES=50
 TRANSACTION_BATCH_SIZE=1
 TRANSACTION_MAX_CONCURRENT_BATCHES=50
-SLOT_UPDATE_INTERVAL_SECONDS=30
\ No newline at end of file
+SLOT_UPDATE_INTERVAL_SECONDS=30
+STATE_QUEUE_START_INDEX=0
+STATE_QUEUE_LENGTH=28807
+ADDRESS_QUEUE_START_INDEX=0
+ADDRESS_QUEUE_LENGTH=28807
```
ananas-block, line 14: how easy is it to change these values?
Would it be easier to change if these are env variables?

ananas-block (approved):
Looks good to me in general. The only thing I am worried about is that right now it seems like we will need to recompile the forester to change the parameters. I think it would be more elegant and flexible to use env variables that we read during runtime so that we can just restart the forester dynamically with a different config in case we want to run multiple foresters on short notice.

PR #1223, opened 2024-09-15
feat: releases + merge-token-accounts method

SwenSchaeferjohann:
* version bumps (as released) 
* another round of minor bumps for now. want to revisit if we have other bigger required changes before going for 1.0.0 on the JS side.
* added method , action, and cli command incl tests for mergeTokenAccounts

* renamed getValidityProof_direct to getValidityProofDirect, un-deprecated it, and fixed it for use with forested queues. this method now allows devs to use a custom prover for the ZKP while using a regular rpc for the merkle proofs


ananas-block (approved):

js/compressed-token/src/idl/light_compressed_token.ts
```diff
@@ -3230,128 +3125,23 @@ export const IDL: LightCompressedToken = {
     errors: [
         {
             code: 6000,
```
sergeytimoshin, line 3127: Is this IDL correct?

js/stateless.js/src/idls/light_compressed_token.ts
```diff
@@ -1549,128 +1549,23 @@ export type LightCompressedToken = {
     errors: [
         {
             code: 6000;
```
sergeytimoshin, line 1551: IDL looks outdated

PR #1239, opened 2024-09-19
fix: pre cargo generate cleanups

ananas-block:


sergeytimoshin (approved):
We should consider refactoring our file helper methods, all of them, in both typescript and rust, are fragile and hacky.

Also: https://github.com/harryfei/which-rs

PR #1254, opened 2024-09-24
feat: insertion circuit

sergeytimoshin:


light-prover/prover/circuit_utils.go
```diff
@@ -115,6 +116,24 @@ func (gadget CombinedProof) DefineGadget(api frontend.API) interface{} {
...
+	currentHash := gadget.Leaf
+	for i := 0; i < len(gadget.Path); i++ {
+		currentHash = abstractor.Call(api, ProveParentHash{
+			Bit:     gadget.Path[i],
+			Hash:    currentHash,
+			Sibling: gadget.Proof[i],
+		})
+	}
+	return currentHash
+}
```
ananas-block, line 135: isn't this the same as `MerkleRootGadget` ?

light-prover/prover/insertion_circuit.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func (p *InsertionParameters) ValidateShape(treeDepth uint32, batchSize uint32) error {
+	if len(p.Leaves) != int(batchSize) {
+		return fmt.Errorf("wrong number of leaves: %d", len(p.Leaves))
+	}
+	if len(p.MerkleProofs) != int(batchSize) {
+		return fmt.Errorf("wrong number of merkle proofs: %d", len(p.MerkleProofs))
+	}
+	for i, proof := range p.MerkleProofs {
```
ananas-block, line 76: I don't think it's necessary to check len of `MerkleProofs` during proofgen because we hardcode the len in the circuit as a setup parameter.

light-prover/prover/insertion_circuit.go
```diff
@@ -0,0 +1,302 @@
...
+			return fmt.Errorf("wrong size of merkle proof for proof %d: %d", i, len(proof))
+		}
+		// for j, element := range proof {
+		// 	if element.Sign() == 0 {
+		// 		return fmt.Errorf("merkle proof element at index [%d][%d] is zero", i, j)
+		// 	}
+		// }
+	}
+
+	// Do we need to check that the leaves are non-zero?
```
ananas-block, line 87: no because we get the leaves from the batch commitment hash onchain.

light-prover/prover/circuit_utils.go
```diff
@@ -222,5 +241,14 @@ func GetKeys(keysDir string, circuitTypes []CircuitType) []string {
 		keys = append(keys, keysDir+"combined_26_4_1.key")
 		keys = append(keys, keysDir+"combined_26_4_2.key")
 	}
+
+	if IsCircuitEnabled(circuitTypes, Insertion) {
+		keys = append(keys, keysDir+"insertion_26_1.key")
+		keys = append(keys, keysDir+"insertion_26_2.key")
+		keys = append(keys, keysDir+"insertion_26_4.key")
+		keys = append(keys, keysDir+"insertion_26_8.key")
```
ananas-block, line 249: I think sensible parameters are 10 for testing, 100 and 500 for actual usage.

light-prover/prover/batch_append_circuit.go
```diff
@@ -0,0 +1,313 @@
...
+		index := api.Add(circuit.StartIndex, i)
+		leaf := circuit.Leaves[i]
+		currentRoot, currentSubtrees = circuit.append(api, leaf, currentSubtrees, index)
+	}
+
+	return currentRoot, currentSubtrees
+}
+
+func (circuit *BatchAppendCircuit) append(api frontend.API, leaf frontend.Variable, subtrees []frontend.Variable, index frontend.Variable) (frontend.Variable, []frontend.Variable) {
+	path := api.ToBinary(index, int(circuit.Depth))
```
ananas-block, line 94: an optimization we can look into once `append` and `update` is merged

`ToBinary` is quite expensive we could try optimize this:
- remove `ToBinary` here
- only call `ToBinary` once per proof
- increment the binary array directly

light-prover/prover/batch_append_circuit.go
```diff
@@ -0,0 +1,313 @@
...
+	HashChainStartIndex uint32     `json:"hashChainStartIndex"`
+}
+
+func (p *BatchAppendParameters) TreeDepth() uint32 {
+	return 26
+}
+
+func (p *BatchAppendParameters) BatchSize() uint32 {
+	return uint32(len(p.Leaves))
+}
```
ananas-block, line 144: note to be careful to keep the circuit general for any tree depth and batch size

light-prover/prover/circuit_utils.go
ananas-block: we can remove it now right?

light-prover/prover/batch_append_circuit.go
```diff
@@ -0,0 +1,364 @@
...
+
+func (circuit *BatchAppendCircuit) getZeroValue(api frontend.API, level int) frontend.Variable {
+	return frontend.Variable(new(big.Int).SetBytes(ZERO_BYTES[level][:]))
+}
+
+func (circuit *BatchAppendCircuit) createHashChain(api frontend.API, length int, hashes []frontend.Variable) frontend.Variable {
+	if length == 0 {
+		return 0
+	}
```
ananas-block, line 180: Shouldn't this be if length == 1 return hashes[0] ?

light-prover/prover/batch_append_circuit.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+func (p *BatchAppendParameters) BatchSize() uint32 {
+	return uint32(len(p.Leaves))
+}
+
+func (p *BatchAppendParameters) ValidateShape(treeDepth uint32, batchSize uint32) error {
+	if p.TreeDepth() != treeDepth {
+		return fmt.Errorf("wrong tree depth: expected %d, got %d", treeDepth, p.TreeDepth())
+	}
```
ananas-block, line 213: you are using `Depth` here while you renamed `Depth` to `Height` in many other places.

light-prover/prover/circuit_builder.go
```diff
@@ -8,20 +8,31 @@ import (
...
-	} else if circuit == NonInclusion {
+	case NonInclusion:
 		return SetupNonInclusion(nonInclusionTreeDepth, nonInclusionNumberOfCompressedAccounts)
-	} else if circuit == Combined {
+	case Combined:
 		return SetupCombined(inclusionTreeDepth, inclusionNumberOfCompressedAccounts, nonInclusionTreeDepth, nonInclusionNumberOfCompressedAccounts)
-	} else {
+	case BatchAppend:
+		return SetupBatchAppend(batchAppendTreeDepth, batchAppendBatchSize)
+	default:
```
ananas-block, line 35: can we do refactors like this in a separate pr pls?

light-prover/prover/batch_append_circuit.go
sergeytimoshin: We covered this case right on the next line https://github.com/Lightprotocol/light-protocol/pull/1254/files/e391cd08c49f91506fd6eb14053e53e6f8dd7e64#diff-41e7c49b3ad64fa70cbeab12a2333eaa7194f8319e9a5f8ee7b1265c86433208R182

light-prover/prover/circuit_builder.go
sergeytimoshin: https://github.com/Lightprotocol/light-protocol/pull/1284
^ let's merge that first, and then I rebase this over new main

light-prover/integration_test.go
```diff
@@ -372,3 +375,72 @@ func TestCombinedHappyPath_JSON(t *testing.T) {
...
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(response.Body)
+		t.Fatalf("Expected status code %d, got %d. Response body: %s", http.StatusOK, response.StatusCode, string(body))
+	}
+}
+
+func TestBatchAppendLargeInput(t *testing.T) {
```
ananas-block, line 399: what's the difference between `TestBatchAppendHappyPath` and `TestBatchAppendLargeInput`?

light-prover/prover/batch_append_circuit_test.go
```diff
@@ -0,0 +1,309 @@
...
+			treeDepth  uint32
+			batchSize  uint32
+			startIndex uint32
+		}{
+			{"Small batch", 26, 10, 0},
+			{"Medium batch", 26, 100, 0},
+			{"Large batch", 26, 1000, 0},
+			{"Tree depth = 4", 4, 10, 0},
+			{"Non-zero start index", 26, 100, 500},
+			{"Start index near tree end", 10, 10, (1 << 10) - 15}, // 2^10 - 15 = 1009
```
ananas-block, line 27: Imo this should not be possible. When integrated with a smart contract which provides this input it is not possible but I am wondering whether we can construct the circuit so that it is also not possible there.
We could for example add next index to the subtree hashchain.

light-prover/scripts/download_keys.sh
```diff
@@ -41,6 +46,39 @@ FILES=(
...
+  "append_4_1.key"
+  "append_4_1.vkey"
+  "append_4_10.key"
+  "append_4_10.vkey"
+  "append_4_100.key"
+  "append_4_100.vkey"
+  "append_4_500.key"
+  "append_4_500.vkey"
+  "append_4_1000.key"
+  "append_4_1000.vkey"
```
ananas-block, line 79: can we remove these?

ananas-block (commented):
Thanks!
Looks good overall, I want to add one more test and execute tests locally before merging this.

light-prover/integration_test.go
sergeytimoshin: Leftover, removed, thanks.

light-prover/prover/batch_append_circuit.go
```diff
@@ -0,0 +1,360 @@
...
+	BatchSize  uint32
+	TreeHeight uint32
+}
+
+func (circuit *BatchAppendCircuit) Define(api frontend.API) error {
+	if err := circuit.validateInputs(); err != nil {
+		return err
+	}
+
+	api.AssertIsEqual(circuit.HashChainStartIndex, 0)
```
ananas-block, line 40: TODOs for a different pr:
- unify all circuits inputs into a single hashchain -> only one public input
- enable partial updates of `HashchainHash`  with `HashChainStartIndex` ( consider `HashchainHash` contains a batch of 5000 leaves but we want to use 5 proves to execute this update onchain)

ananas-block (approved):
Looks good to me, once the TODOs I specified above are added and its rebased.

light-prover/prover/batch_append_circuit.go
ananas-block:
```suggestion
TODO:
	- unify all circuits inputs into a single hashchain -> only one public input
	- enable partial updates of `HashchainHash`  with `HashChainStartIndex`
	( consider `HashchainHash` contains a batch of 5000 leaves but we want to use 5 proves to execute this update onchain)
	api.AssertIsEqual(circuit.HashChainStartIndex, 0)
```

ananas-block (approved):

PR #1260, opened 2024-09-25
feat: batch update circuit

sergeytimoshin:


circuit-lib/verifier/src/verifying_keys/combined_26_1_1.rs
ananas-block, line 1: we shouldn't regenerate vks for existing circuits since these are based on the trusted setup

light-prover/prover/update_circuit.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/reilabs/gnark-lean-extractor/v2/abstractor"
+)
+
+type BatchUpdateCircuit struct {
+	PreRoot    frontend.Variable `gnark:",public"`
+	PostRoot   frontend.Variable `gnark:",public"`
+	StartIndex frontend.Variable `gnark:"input"`
+
+	OldLeaves    []frontend.Variable   `gnark:"input"`
+	NewLeaves    []frontend.Variable   `gnark:"input"`
```
ananas-block, line 21: Note: `NewLeaves` will always be 0 for state Merkle tree updates. I can see merit in making the circuit generic. However, we should make sure that dynamic `NewLeaves` don't come at a performance penalty.
Have you benchmarked dynamic `NewLeaves` vs hardcoding an update with '0' in circuit?

light-prover/prover/update_circuit.go
```diff
@@ -0,0 +1,296 @@
...
+			Leaf:  gadget.OldLeaves[i],
+			Path:  api.ToBinary(currentIndex, int(gadget.Depth)),
+			Proof: gadget.MerkleProofs[i],
+		})
+		api.AssertIsEqual(oldRoot, currentRoot)
+
+		// Update with the new leaf
+		newRoot := abstractor.Call(api, VerifyProof{
+			Leaf:  gadget.NewLeaves[i],
+			Path:  api.ToBinary(currentIndex, int(gadget.Depth)),
```
ananas-block, line 212: possible optimization just execute `ToBinary` once.

light-prover/prover/insertion_circuit.go
ananas-block, line 1: can we rename this to append?

light-prover/prover/insertion_circuit.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/frontend/cs/r1cs"
+	"github.com/reilabs/gnark-lean-extractor/v2/abstractor"
+)
+
+type InsertionCircuit struct {
+	// InputHash frontend.Variable `gnark:",public"`
+
+	// private inputs
+	StartIndex frontend.Variable `gnark:"input"`
```
ananas-block, line 21: wondering whether start index should be public as well.

PR #1261, opened 2024-09-25
fix: macos linker fix

SwenSchaeferjohann:


vadorovsky (approved):
Looks good, but does it also fix the "could not find lld" problem? My guess is that it doesn't, since macOS doesn't ship it by default.

I was thinking about another solution - we could just download binary tarballs of LLVM from here:

https://github.com/llvm/llvm-project/releases/tag/llvmorg-19.1.0

PR #1291, opened 2024-10-11
feat: batch update circuit

sergeytimoshin:


light-prover/prover/batch_update_circuit.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/consensys/gnark/constraint"
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/frontend/cs/r1cs"
+	"github.com/reilabs/gnark-lean-extractor/v2/abstractor"
+)
+
+type BatchUpdateCircuit struct {
+	OldRoot             frontend.Variable `gnark:",public"`
+	NewRoot             frontend.Variable `gnark:",public"`
+	HashChainStartIndex frontend.Variable `gnark:",public"`
```
ananas-block, line 20: you can remove this I think it doesn't make sense afterall.
Instead we will require the number of batched zkp updates it requires to insert a batch into the tree to be fixed at tree initialization and create a separate LeavesHashchainHash for each zkp batch.

light-prover/prover/batch_update_circuit.go
```diff
@@ -0,0 +1,256 @@
...
+		api.AssertIsEqual(currentRoot, newRoot)
+		newRoot = circuit.merkleRoot(api, emptyLeaf, indexBits, circuit.MerkleProofs[i])
+	}
+
+	api.AssertIsEqual(newRoot, circuit.NewRoot)
+
+	return nil
+}
+
+func (circuit *BatchUpdateCircuit) merkleRoot(api frontend.API, leaf frontend.Variable, indexBits []frontend.Variable, siblings []frontend.Variable) frontend.Variable {
```
ananas-block, line 52: can we unify this with the verify Merkle proof implementation in circuit_utils?

light-prover/prover/batch_update_circuit.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (circuit *BatchUpdateCircuit) incrementBits(api frontend.API, bits []frontend.Variable) []frontend.Variable {
+	carry := frontend.Variable(1)
+	for i := 0; i < len(bits); i++ {
+		newBit := api.Xor(bits[i], carry)
+		carry = api.And(bits[i], carry)
+		bits[i] = newBit
+	}
+	return bits
+}
```
ananas-block, line 72: I believe this in not used here right?

light-prover/prover/batch_update_circuit.go
```diff
@@ -0,0 +1,256 @@
...
+	if length == 0 {
+		return frontend.Variable(0)
+	}
+
+	hashChain := hashes[0]
+	for i := 1; i < length; i++ {
+		hashChain = abstractor.Call(api, poseidon.Poseidon2{In1: hashChain, In2: hashes[i]})
+	}
+	return hashChain
+}
```
ananas-block, line 84: pls unify

circuit-lib/verifier/src/verifying_keys/update_10_500.rs
```diff
@@ -0,0 +1,79 @@
+// This file is generated by xtask. Do not edit it manually.
+
+use groth16_solana::groth16::Groth16Verifyingkey;
+pub const VERIFYINGKEY: Groth16Verifyingkey = Groth16Verifyingkey {
```
ananas-block, line 4: just vks for height 26 for both append and update pls

light-prover/prover/batch_update_circuit.go
ananas-block: `MerkleRootGadget` in `circuit_utils` is equivalent with this gadget right?

sergeytimoshin: practically yes, I'll unify it

ananas-block (approved):

PR #1294, opened 2024-10-12
chore: add missing install reqirement

ananas-block:
Changes:
- add missing requirements to readme
- taken from [config.toml comments](https://github.com/Lightprotocol/light-protocol/blob/main/.cargo/config.toml#L4-L30)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,12 +70,12 @@ There are three ways of setting up the development environment:
   unfortunately has performance problems on macOS.
 - [Manual setup](#manual-setup) - not recommended, but may be useful if the
   methods above don't work for you.
+- Windows is not supported
 
 ### Prerequisites:
 - Ubuntu, `sudo apt-get install lld clang`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,6 +71,12 @@ There are three ways of setting up the development environment:
 - [Manual setup](#manual-setup) - not recommended, but may be useful if the
   methods above don't work for you.
 
+### Prerequisites:
+- Ubuntu, `sudo apt-get install lld clang`
+- Arch, `sudo pacman -S lld clang`
+- Mac, `brew install llvm`
+- Windows `cargo install -f cargo-binutils && rustup component add llvm-tools-preview`
```
sergeytimoshin, line 78: I think we should explicitly state that we do not support Windows (at least for now).

sergeytimoshin (changes requested):

sergeytimoshin (approved):

PR #1295, opened 2024-10-12
refactor: prover: support lightweight integration tests

sergeytimoshin:


light-prover/prover/circuit_utils.go
```diff
@@ -222,38 +222,53 @@ func LoadVerifyingKey(filepath string) (verifyingKey groth16.VerifyingKey, err e
...
+		} else {
+			keys = append(keys, keysDir+"combined_26_1_1.key")
+			keys = append(keys, keysDir+"combined_26_1_2.key")
+			keys = append(keys, keysDir+"combined_26_2_1.key")
+			keys = append(keys, keysDir+"combined_26_2_2.key")
+			keys = append(keys, keysDir+"combined_26_3_1.key")
+			keys = append(keys, keysDir+"combined_26_3_2.key")
+			keys = append(keys, keysDir+"combined_26_4_1.key")
+			keys = append(keys, keysDir+"combined_26_4_2.key")
+		}
```
ananas-block, line 259: this might brick program and js tests, maybe add another flag for program tests which downloads all of these?

scripts/install.sh
```diff
@@ -187,14 +187,29 @@ install_dependencies() {
...
+            --full-keys)
+                key_type="full"
+                shift
+                ;;
+            *)
+                echo "Unknown option: $1"
+                exit 1
+                ;;
+        esac
+    done
```
ananas-block, line 203: for all other ci tests than prover, do we currently download all provingkeys?

sergeytimoshin: No, by default we download only subset of proving keys just for tests, I wrote a note about it here https://github.com/Lightprotocol/light-protocol/pull/1295/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R86-R91

js/stateless.js/src/test-helpers/test-rpc/test-rpc.ts
```diff
@@ -142,7 +142,7 @@ export class TestRpc extends Connection implements CompressionApiInterface {
     addressQueueAddress: PublicKey;
     lightWasm: LightWasm;
     depth: number;
-    log = false;
+    log = true;
```
ananas-block, line 145: is this intentional?

PR #1300, opened 2024-10-14
chore: add comments about how append method works

sergeytimoshin:


light-prover/prover/batch_append_circuit.go
```diff
@@ -88,6 +88,34 @@ func (circuit *BatchAppendCircuit) batchAppend(api frontend.API) (frontend.Varia
...
+//     a. It uses the corresponding bit of the index to determine if we're inserting on the right or left.
+//     b. If inserting on the right (isRight is true):
+//     - The current subtree at this level becomes the left sibling.
+//     - The current node becomes the right sibling and the new subtree for this level.
+//     c. If inserting on the left (isRight is false):
+//     - The current node becomes the left sibling.
+//     - The existing subtree at this level (or a zero value if none exists) becomes the right sibling.
+//     d. It selects the appropriate left and right nodes based on the insertion direction.
+//     e. It hashes the left and right nodes together to create the new parent node.
+//     f. This new parent becomes the current node for the next level up.
```
ananas-block, line 107:
```suggestion
//     b. update subtrees:
//     - if isRight then do not update subtrees
//     - else update subtrees with current node
//     c. If inserting on the right (isRight is true):
//     - The current subtree at this level becomes the left sibling.
//     - The current node becomes the right sibling and the new subtree for this level.
//     d. If inserting on the left (isRight is false):
//     - The current node becomes the left sibling.
//     -  zero value becomes the right sibling.
//     e. It selects the appropriate left and right nodes based on the insertion direction.
//     f. It hashes the left and right nodes together to create the new parent node.
//     g. This new parent becomes the current node for the next level up.
```

ananas-block (approved):

PR #1304, opened 2024-10-16
fix: add light prover client batch update and append tests

ananas-block:
Changes:
- add tests for `light-prover-client` batch update and append

### DIFF
--- a/light-prover/main.go
+++ b/light-prover/main.go
@@ -791,10 +791,6 @@ func getKeysByArgs(context *cli.Context, isTestMode bool) ([]string, error) {
 		circuitTypes = append(circuitTypes, prover.Combined)
 	}
 
-	if !inclusion && !nonInclusion {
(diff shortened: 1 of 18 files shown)


light-prover/main.go
```diff
@@ -791,10 +791,6 @@ func getKeysByArgs(context *cli.Context, isTestMode bool) ([]string, error) {
 		circuitTypes = append(circuitTypes, prover.Combined)
 	}
 
-	if !inclusion && !nonInclusion {
-		return nil, fmt.Errorf("no circuit type provided")
-	}
```
ananas-block, line 796: this prevented the prover from starting with only append or update circuits.

sergeytimoshin: fix https://github.com/Lightprotocol/light-protocol/pull/1305

sergeytimoshin (approved):

PR #1332, opened 2024-11-09
feat: getCompressedMintTokenholders endpoint + JS release bump

SwenSchaeferjohann:
* added: rpc endpoint getCompressedMintTokenholders 
* bumped photon to 0.47.0
* released bumped cli + js sdks
* added: test coverage for paginated response 
* fixed: light test-validator without providing flags, properly fallback to default flags (would force flags before)
* fixed: removed anchor provider and airdrop at startup (would silently crash on mac before)


cli/.npmignore
```diff
@@ -1 +1,7 @@
-bin/cargo-generate
\ No newline at end of file
+
```
sergeytimoshin, line 1: I would prefer to ignore bin/proving-keys folder instead of files inside (this list of files is not comprehensive, if you download all proving keys via `/light-prover/scripts/dowload_keys.sh full`, this list would be a much longer.

sergeytimoshin (approved):

PR #1345, opened 2024-11-15
chore: standalone photonclient

SwenSchaeferjohann:


ananas-block (commented):
can we add a test for the client?

sergeytimoshin (approved):

PR #1354, opened 2024-11-20
feat: add light-program-test crate

sergeytimoshin:


test-utils/src/test_env.rs
```diff
@@ -418,7 +418,7 @@ pub async fn setup_test_programs_with_accounts(
 pub async fn setup_test_programs_with_accounts_v2(
     additional_programs: Option<Vec<(String, Pubkey)>>,
 ) -> (
-    light_client::rpc::test_rpc::ProgramTestRpcConnection,
+    light_program_test::test_rpc::ProgramTestRpcConnection,
```
ananas-block, line 421: can you move this file to light-program-test as well since it creates the test env which the test indexer is based on?

sergeytimoshin:
move to https://github.com/Lightprotocol/light-protocol/pull/1372

PR #1357, opened 2024-11-22
feat: forester batched ops

sergeytimoshin:


programs/account-compression/src/state/batched_queue.rs
```diff
@@ -0,0 +1,827 @@
...
+    ) -> Result<()> {
+        for (batch_index, batch) in self.batches.iter().enumerate() {
+            if batch.value_is_inserted_in_batch(leaf_index)? {
+                let index = batch.get_value_index_in_batch(leaf_index)?;
+                let element = self.value_vecs[batch_index]
+                    .get_mut(index as usize)
+                    .ok_or(AccountCompressionErrorCode::InclusionProofByIndexFailed)?;
+
+                if element == value {
+                    *element = [0; 32];
```
ananas-block, line 274:
```suggestion
```
We need to remove this so that we can use the value from `output_queue.value_vec[full_batch_index]` directly instead from the indexer.

forester/src/batched_ops.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        let (old_leaves, merkle_proofs) = {
+            let mut old_leaves = vec![];
+            let mut merkle_proofs = vec![];
+            for i in merkle_tree_next_index..merkle_tree_next_index + zkp_batch_size {
+                let proof = self
+                    .indexer
+                    .lock()
+                    .await
+                    .get_proof_by_index(self.merkle_tree, i as usize);
```
ananas-block, line 263: should we add a batched call to the indexer for this?

forester/src/batched_ops.rs
```diff
@@ -0,0 +1,481 @@
...
+            .get_leaf_indices_tx_hashes(self.merkle_tree, zkp_batch_size as usize);
+
+        let mut leaves = Vec::new();
+        let mut tx_hashes = Vec::new();
+        let mut old_leaves = Vec::new();
+        let mut path_indices = Vec::new();
+        let mut merkle_proofs = Vec::new();
+
+        let batch = {
+            let mut merkle_tree_account = rpc.get_account(self.merkle_tree).await.unwrap().unwrap();
```
ananas-block, line 360: duplicate `get_account` call its already fetched in line 331.

forester/src/batched_ops.rs
```diff
@@ -0,0 +1,481 @@
...
+            let root = *merkle_tree.root_history.last().unwrap();
+
+            (zkp_size, batch_idx, root, root_idx, hashchain)
+        };
+
+        let leaf_indices_tx_hashes = self
+            .indexer
+            .lock()
+            .await
+            .get_leaf_indices_tx_hashes(self.merkle_tree, zkp_batch_size as usize);
```
ananas-block, line 351: wondering whether it would make sense to return `get_proof_by_index(self.merkle_tree, *index as usize);` with this call as well so that we avoid the second call to the indexer.

programs/account-compression/src/state/batched_queue.rs
```diff
@@ -271,7 +267,6 @@ impl ZeroCopyBatchedQueueAccount {
                     .ok_or(AccountCompressionErrorCode::InclusionProofByIndexFailed)?;
 
                 if element == value {
-                    *element = [0; 32];
```
ananas-block, line 274: unfortunately we need to revert this because it leads to some issues onchain.
This means that we can't get the leaves for the batch append circuit inputs from the batch Merkle tree account value array because some values have been zeroed out.
Instead we need to get the data from the indexer. I think it makes sense to add a `get_queue_elements` method to the `Indexer` trait like I have for the batch address tree [see here](https://github.com/Lightprotocol/light-protocol/pull/1365/commits/4df70ae882f9cfe8e93992b4cf4e9522cbf8d489#diff-42faea6ff4f47f30e026b0ead2044d53a053289517ed9aa60ae996efaa87f30dR76-R91).

PR #1361, opened 2024-11-25
feat: compress spl token account

ananas-block:
### Changes:
- add instruction to compress a spl token account with an optional remaining amount
- `process_compress_spl_token_account`
    - creates an `compressed_output_account`
    - wraps `process_transfer`
- moved instruction data deserialization out of  `process_transfer` so that `process_compress_spl_token_account` can wrap it

### DIFF
--- a/programs/compressed-token/src/lib.rs
+++ b/programs/compressed-token/src/lib.rs
@@ -1,8 +1,10 @@
 use anchor_lang::prelude::*;
 
 pub mod constants;
+pub mod process_compress_spl_token_account;
 pub mod process_mint;
 pub mod process_transfer;
+use process_compress_spl_token_account::process_compress_spl_token_account;
 pub mod spl_compression;
 pub use process_mint::*;
 pub mod token_data;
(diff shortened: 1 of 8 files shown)


SwenSchaeferjohann (approved):

programs/compressed-token/src/lib.rs
```diff
@@ -57,6 +60,15 @@ pub mod light_compressed_token {
         process_mint_to(ctx, public_keys, amounts, lamports)
     }
```
robre, line 62: Should add a docstring to this function, as for the others in this file.

programs/compressed-token/src/process_compress_spl_token_account.rs
```diff
@@ -0,0 +1,107 @@
...
+    let compression_token_account =
+        if let Some(token_account) = ctx.accounts.compress_or_decompress_token_account.as_ref() {
+            token_account
+        } else {
+            return err!(ErrorCode::CompressedPdaUndefinedForCompress);
+        };
+    let compress_amount = compression_token_account
+        .amount
+        .checked_sub(remaining_amount.unwrap_or_default())
+        .ok_or(ErrorCode::ArithmeticUnderflow)?;
```
robre, line 26: The ErrorCode could be more precise,
Users who put in a remaining_amount that's higher than the available amount should maybe get an error that states insufficient funds or something like that

PR #1362, opened 2024-11-25
feat: add JS support for compress-spl-token-account instruction

SwenSchaeferjohann:
based on #1361


js/compressed-token/src/program.ts
```diff
@@ -991,4 +1019,70 @@ export class CompressedTokenProgram {
...
+        // const {
+        //     inputTokenDataWithContext,
+        //     packedOutputTokenData,
+        //     remainingAccountMetas,
+        // } = packCompressedTokenAccounts({
+        //     inputCompressedTokenAccounts: [],
+        //     outputStateTrees: outputStateTree,
+        //     rootIndices: [],
+        //     tokenTransferOutputs,
+        // });
```
ananas-block, line 1053: leftover?

PR #1364, opened 2024-11-27
chore: sdk w/o anchor

SwenSchaeferjohann:
based on #1349


sdk/src/constants.rs
```diff
@@ -4,6 +4,8 @@ use solana_program::pubkey::Pubkey;
 /// Seed of the CPI authority.
 pub const CPI_AUTHORITY_PDA_SEED: &[u8] = b"cpi_authority";
 
+/// ID of the Solana system program.
+pub const PROGRAM_ID_SYSTEM: Pubkey = pubkey!("11111111111111111111111111111111");
```
ananas-block, line 8: you can just use `Pubkey::default()`

sdk/src/account_meta.rs
```diff
@@ -1,79 +1,193 @@
...
+//             output_merkle_tree_index: None,
+//             address_merkle_context: None,
+//             address_merkle_tree_root_index: None,
+//             read_only: false,
+//         }
+//     }
+// }
+
+#[derive(Debug, Clone, PartialEq)]
 pub struct LightAccountMeta {
```
ananas-block, line 122: should we add a type field (enum Init, Mut, ReadOnly, Close) which is used to pack the AccountMeta?

PR #1365, opened 2024-11-28
feat: batch address mt

ananas-block:


programs/system/src/sdk/address.rs
```diff
@@ -15,6 +18,21 @@ pub fn derive_address(merkle_tree_pubkey: &Pubkey, seed: &[u8; 32]) -> Result<[u
     Ok(hash)
 }
 
+pub fn derive_address(
```
ananas-block, line 21: - don't hash program id since we hash it with keccak anyway

programs/system/src/invoke/address.rs
```diff
@@ -1,37 +1,77 @@
...
     compressed_account_addresses: &mut [Option<[u8; 32]>],
-    new_addresses: &mut [[u8; 32]],
+    new_addresses: &mut Vec<[u8; 32]>,
 ) -> Result<()> {
+    let hashed_invoking_program_id = if let Some(invoking_program_id) = invoking_program_id {
+        Some(
+            hash_to_bn254_field_size_be(&invoking_program_id.to_bytes())
+                .ok_or(SystemProgramError::DeriveAddressError)?
+                .0,
+        )
```
ananas-block, line 32: remove it's not necessary to hash the program ID because the address hash is a truncated keccak hash anyway

PR #1373, opened 2024-12-03
refactor: move test_env to light-program-test

sergeytimoshin:


light-program-test/src/asserts/assert_address_merkle_tree.rs
ananas-block, line 1: can we leave asserts in `test-utils`?
so that we have everything that is interesting for external devs in `light-program-test` and everything that we only need internally remains in `light_test_utils`.

ananas-block: ah I see why you moved them the perform actions use the assert that's ugly.
The ideal would be to wrappers in test-utils which are `perform_..._with_assert` that we can keep using in our tests and the `E2eTestEnv`. I think `E2eTestEnv` should also remain in `light-test-utils` to keep `light-program-test` as slim as possible imo.

sergeytimoshin: > The ideal would be to wrappers in test-utils which are perform_..._with_assert that we can keep using in our tests and the E2eTestEnv. 

This is the good solution for 99% percent of our cases.
1. The only place in `light-program-test` where we use asserts is in the  [`create_address_merkle_tree_and_queue_account`](https://github.com/Lightprotocol/light-protocol/blob/sergey/move-test_env-to-program-test/light-program-test/src/test_env.rs#L840) function, and we can wrap it in something like `perform_create_address_merkle_tree_and_queue_account_with_assert` and put it in `test-utils`. 
2. The problem is that `test_indexer` also [uses this function](https://github.com/Lightprotocol/light-protocol/blob/sergey/move-test_env-to-program-test/test-utils/src/indexer/test_indexer.rs#L697), and if/when we move indexer to `light-program-test` it wouldnt be possible to use the wrapped variant inside the indexer anymore.

ananas-block: re 2: if we extract the assert into a wrapper then we could keep using it right?

test-utils/src/lib.rs
```diff
@@ -9,17 +15,16 @@ pub mod assert_queue;
 pub mod assert_rollover;
 pub mod assert_token_tx;
 pub mod e2e_test_env;
-pub mod env_accounts;
 #[allow(unused)]
 pub mod indexer;
```
ananas-block, line 19: can we remove `indexer.rs` as well?

ananas-block: saw that its in a different pr

ananas-block (approved):

PR #1376, opened 2024-12-06
refactor: test indexer

sergeytimoshin:
1. TestIndexer implementations merged and [moved to light-program-test](https://github.com/Lightprotocol/light-protocol/blob/sergey/refactor_test_indexer/light-program-test/src/test_indexer.rs).
2. [Added helper functions](https://github.com/Lightprotocol/light-protocol/blob/sergey/refactor_test_indexer/test-utils/src/conversions.rs) to convert data types between programs & sdk.


client/Cargo.toml
```diff
@@ -34,12 +34,16 @@ log = { workspace = true }
 thiserror = { workspace = true }
 
 # Light Protocol
+account-compression = { workspace = true }
+light-compressed-token = { workspace = true }
```
ananas-block, line 38: pls remove these dependencies we would like to keep `light-client`, `light-sdk` etc free of anchor hence program dependencies.

client/src/indexer/mod.rs
```diff
@@ -1,57 +1,172 @@
...
+    pub address_root_indices: Vec<u16>,
+}
+
+#[derive(Debug, Clone)]
+pub struct TokenDataWithMerkleContext {
+    pub token_data: TokenData,
+    pub compressed_account: CompressedAccountWithMerkleContext,
 }
 
 pub trait Indexer<R: RpcConnection>: Sync + Send + Debug + 'static {
```
ananas-block, line 62: do we even need the `Indexer` trait can't we just have a `photon-indexer` trait and a `test-indexer` trait?

forester/src/send_transaction.rs
```diff
@@ -395,8 +395,9 @@ pub async fn fetch_proofs_and_create_instructions<R: RpcConnection, I: Indexer<R
         join!(address_future, state_future)
     };
 
-    let address_proofs = address_proofs?;
-    let state_proofs = state_proofs?;
+    // TODO: handle errors
+    let address_proofs = address_proofs.unwrap();
+    let state_proofs = state_proofs.unwrap();
```
ananas-block, line 400:
```suggestion
    let state_proofs = state_proofs.ok_or(ErrorCode::RandomError)?;
```

client/Cargo.toml
ananas-block: seeing that you are importing struct definitions from the program dependencies.
`light-sdk` contains many copy pasted struct definitions already to avoid program dependencies. You can probably import most from there. In case some are missing fee free to add them in `light-client` as well for example for `PublicTransactionEvent` although I am not sure that you need that outside of the `test-indexer`.

sergeytimoshin: > else copy pasta struct definition where you need it

I don't like this approach tbh, its error-prone and hard to handle, especially in `test-utils` and `test-programs`, where every second test uses structs from programs and `light-sdk` / `client` simultaneously, leading to nasty code like this https://github.com/Lightprotocol/light-protocol/blob/sergey/refactor_test_indexer/test-programs/sdk-test-program/programs/sdk-test/tests/test.rs#L99-L129

wouldn't it be better to extract these common structs into separate crate, and use this crate in `*-programs` and in `client`& `sdk`?

ananas-block: **I see 4 options here the tldr see for details below:**
1. upgrade anchor and extract programs, don't like it a lot of effort and will remove anchor next year
2. introduce an sdk feature to programs and extract the sdk with and without anchor
3. extract structs into one crate but just copy paste don't use this in programs
4. copy pasta as Michal did it for the sdk

leaning towards 3 or 4, @sergeytimoshin wdyt?

**Details:**
1. The reason why we haven't extracted the types so far is that 0.29.0 idl generation doesn't work anymore and bumping to anchor 0.30.0 has proven to be a nightmare. I don't think this is a solution, I want to ditch anchor completely next year to optimize our core programs.

2. Maybe an easier way is to make anchor an optional dependency for programs and create a feature which only compiles the system-program/src/sdk/* then we can import types from programs without the anchor dependency.
- add feature sdk
- make anchor optional
- import programs to sdk without default features
This could be a fast and non-intrusive way but it will increase compile time of the sdk with anchor a lot which I don't like.

3. Different way, create one crate which contains all the relevant extracted structs with copy pasta and use it. Don't import programs  Still copy pasta but everything is in one place.
    - easier to maintain than copy pasta in multiple places
    - doesn't increase compile time of the sdk

4. Or just stick to copy pasta for now, the system program instruction bindings are not going to change ever with it the structs to generate instruction data. We are just going to add new instructions going forward and possibly deprecate old once.

sergeytimoshin: Hm.
I like both approaches 3 (separate crate) and 4 (copy-paste), but there's still the problem of manual type conversions in tests like:
```
let mut sdk_compressed_account_data: Option<CompressedAccountData> = None;
if let Some(compressed_account_data) = &compressed_account.compressed_account.data {
    sdk_compressed_account_data = Some(CompressedAccountData {
        discriminator: compressed_account_data.discriminator,
        // ... lots of manual field mapping
    });
}
```

Instead of choosing between 3 and 4, I think about hybrid approach:

1. Keep copy-pasting structs in the SDK (approach 4)
2. Add conversion traits ONLY in test code using cfg(test), allowing ergonomic conversions like:
`let sdk_account: sdk::CompressedAccountWithMerkleContext = program_account.into();`

This gives us:
1. Clean SDK without program dependencies
2. A lot less pain wrt type conversion in tests
3. Still easy path to remove Anchor later

ananas-block: sounds good but please make anchor optional from the start.
Michal found an easy way see:
https://github.com/Lightprotocol/light-protocol/pull/1364/files#diff-7972129256fe1dd3b281558ba8f49d1540d7c1cafb3b2033f1335a0a83645708R1-R4

sergeytimoshin: No, we can't put conversion traits in `test-utils` because of [E0117 (Only traits defined in the current crate can be implemented for arbitrary types.)](https://doc.rust-lang.org/error_codes/E0117.html).

This it what I proposing, just one conversion as an example: https://github.com/Lightprotocol/light-protocol/pull/1376/commits/7c5c7e66ff4c6f88f30dd5072db2581bb94e3372

If you ok with it, I will do the rest conversions like this one.

ananas-block: the issue is that this requires sdk to import the programs which we wanted to avoid.

sergeytimoshin: Yes, correct

client/src/rpc/solana_rpc.rs
```diff
@@ -434,3 +435,5 @@ impl RpcConnection for SolanaRpcConnection {
         .await
     }
 }
+
+impl MerkleTreeExt for SolanaRpcConnection {}
```
ananas-block, line 439: feels odd to me, could we get away without this?

client/src/indexer/mod.rs
ananas-block: but indexers don't have to support methods like `get_address_merkle_trees_mut` these are `test-indexer` specific.
Essentially every method which is `unimplemented` by default can probably be moved to the test-indexer trait so that we get rid of all the `unimplemented` methods.
Do you think that would work?

test-programs/sdk-test-program/programs/sdk-test/tests/test.rs
```diff
@@ -168,8 +171,18 @@ where
     )
     .unwrap();
 
+    let proof_result = ProofRpcResult {
+        root_indices: rpc_result.address_root_indices.clone(),
```
ananas-block, line 175: can't we just use one `ProofRpcResult` type in light-client everywhere, why do we need two?

client/src/indexer/mod.rs
ananas-block: These are the methods I have in mind. Some already exist in the `MerkleTreeExt` trait already.
```
fn account_nullified(&mut self, _merkle_tree_pubkey: Pubkey, _account_hash: &str) {}

    fn address_tree_updated(
        &mut self,
        _merkle_tree_pubkey: Pubkey,
        _context: &NewAddressProofWithContext,
    ) {
    }

    fn get_state_merkle_tree_accounts(&self, _pubkeys: &[Pubkey]) -> Vec<StateMerkleTreeAccounts> {
        unimplemented!()
    }

    fn add_event_and_compressed_accounts(
        &mut self,
        _event: &PublicTransactionEvent,
    ) -> (
        Vec<CompressedAccountWithMerkleContext>,
        Vec<TokenDataWithMerkleContext>,
    ) {
        unimplemented!()
    }

    fn get_state_merkle_trees(&self) -> &Vec<StateMerkleTreeBundle> {
        unimplemented!()
    }

    fn get_state_merkle_trees_mut(&mut self) -> &mut Vec<StateMerkleTreeBundle> {
        unimplemented!()
    }

    fn get_address_merkle_trees(&self) -> &Vec<AddressMerkleTreeBundle> {
        unimplemented!()
    }

    fn get_address_merkle_trees_mut(&mut self) -> &mut Vec<AddressMerkleTreeBundle> {
        unimplemented!()
    }

    fn get_token_compressed_accounts(&self) -> &Vec<TokenDataWithMerkleContext> {
        unimplemented!()
    }

    fn get_payer(&self) -> &Keypair {
        unimplemented!()
    }

    fn get_group_pda(&self) -> &Pubkey {
        unimplemented!()
    }

    #[allow(async_fn_in_trait)]
    async fn create_proof_for_compressed_accounts(
        &mut self,
        _compressed_accounts: Option<&[[u8; 32]]>,
        _state_merkle_tree_pubkeys: Option<&[Pubkey]>,
        _new_addresses: Option<&[[u8; 32]]>,
        _address_merkle_tree_pubkeys: Option<Vec<Pubkey>>,
        _rpc: &mut R,
    ) -> ProofRpcResult {
        unimplemented!()
    }

    fn add_address_merkle_tree_accounts(
        &mut self,
        _merkle_tree_keypair: &Keypair,
        _queue_keypair: &Keypair,
        _owning_program_id: Option<Pubkey>,
    ) -> AddressMerkleTreeAccounts {
        unimplemented!()
    }

    fn get_compressed_accounts_by_owner(
        &self,
        _owner: &Pubkey,
    ) -> Vec<CompressedAccountWithMerkleContext> {
        unimplemented!()
    }

    fn get_compressed_token_accounts_by_owner(
        &self,
        _owner: &Pubkey,
    ) -> Vec<TokenDataWithMerkleContext> {
        unimplemented!()
    }

    fn add_state_bundle(&mut self, _state_bundle: StateMerkleTreeBundle) {
        unimplemented!()
    }
    ```

File: test-programs/sdk-test-program/programs/sdk-test/tests/test.rs (hunk 5 above)
Comment (sergeytimoshin) [reply test-programs/sdk-test-program/programs/sdk-test/tests/test.rs:175]: We have only one `ProofRpcResult`, [its declared in `sdk`](https://github.com/Lightprotocol/light-protocol/blob/sergey/refactor_test_indexer/sdk/src/proof.rs#L37-L42).

File: client/src/indexer/mod.rs [hunk 6]
```diff
@@ -1,57 +1,31 @@
...
-        address_merkle_tree_pubkeys: Option<Vec<Pubkey>>,
-        rpc: &mut R,
-    ) -> impl Future<Output = ProofRpcResult>;
+    fn get_multiple_compressed_account_proofs(
+        &self,
+        hashes: Vec<String>,
+    ) -> impl std::future::Future<Output = Result<Vec<MerkleProof>, IndexerError>> + Send + Sync;
 
-    fn get_compressed_accounts_by_owner(
+    fn get_rpc_compressed_accounts_by_owner(
```
ananas-block, line 17: why rename this?

test-programs/system-test/tests/test.rs
```diff
@@ -80,6 +83,7 @@ async fn invoke_failing_test() {
 
     let payer = context.get_payer().insecure_clone();
     // no inputs
+    println!("!!! 1");
```
ananas-block, line 86: leftover

test-utils/src/e2e_test_env.rs
```diff
@@ -1947,17 +1952,48 @@ where
...
             .indexer
-            .get_compressed_accounts_by_owner(&self.users[user_index].keypair.pubkey());
+            .get_compressed_accounts_by_owner(&self.users[user_index].keypair.pubkey())
+            .into_iter()
+            .map(sdk_to_program_compressed_account_with_merkle_context)
+            .collect::<Vec<_>>();
         let range = std::cmp::min(input_compressed_accounts.len(), 4);
         let number_of_compressed_accounts = Self::safe_gen_range(&mut self.rng, 0..=range, 0);
         input_compressed_accounts[0..number_of_compressed_accounts].to_vec()
+        // TODO: convert
```
ananas-block, line 1962: does this TODO still make sense?

ananas-block (approved):

PR #1388, opened 2024-12-11
feat: add --geyser-config and --validator-args CLI flags

SwenSchaeferjohann:
+ released as cli 0.22.0


ananas-block (commented):
Is versioning correct?

package.json is `0.22.0` while pr comment says `released as cli 0.20.0`

ananas-block (approved):

PR #1393, opened 2024-12-14
feat: add support for `--sbf-program` flag in test-validator

sergeytimoshin:
This allows adding SBF programs to the genesis configuration with upgrades disabled. The flag accepts pairs of arguments (address and path) to specify programs, and validates input to ensure correct usage.

### DIFF
--- a/cli/src/commands/test-validator/index.ts
+++ b/cli/src/commands/test-validator/index.ts
@@ -17,6 +17,7 @@ class SetupCommand extends Command {
     "$ light test-validator --skip-indexer",
     "$ light test-validator --geyser-config ./config.json",
     '$ light test-validator --validator-args "--limit-ledger-size 50000000"',
+    "$ light test-validator --sbf-program <address> <path/program>",
   ];
 
   protected finally(err: Error | undefined): Promise<any> {
--- a/cli/src/commands/test-validator/index.ts
+++ b/cli/src/commands/test-validator/index.ts
@@ -1,5 +1,9 @@
 import { Command, Flags } from "@oclif/core";
-import { initTestEnv, stopTestEnv } from "../../utils/initTestEnv";
+import {
+  initTestEnv,
+  stopTestEnv,
+  SYSTEM_PROGRAMS,
+} from "../../utils/initTestEnv";
 import { CustomLoader } from "../../utils/index";
 import path from "path";
 import fs from "fs";
@@ -15,7 +19,10 @@ class SetupCommand extends Command {
     '$ light test-validator --validator-args "--limit-ledger-size 50000000"',
   ];
 
-  protected finally(_: Error | undefined): Promise<any> {
+  protected finally(err: Error | undefined): Promise<any> {
+    if (err) {
+      console.error(err);
+    }
     process.exit();
   }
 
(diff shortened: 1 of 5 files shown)


cli/src/commands/test-validator/index.ts
```diff
@@ -119,6 +119,12 @@ class SetupCommand extends Command {
       required: false,
       exclusive: ["geyser-config"],
     }),
+    "sbf-program": Flags.string({
+      description:
+        "Add a SBF program to the genesis configuration with upgrades disabled. If the ledger already exists then this parameter is silently ignored. First argument can be a pubkey string or path to a keypair",
```
ananas-block, line 124:
```suggestion
        "<address> <path> Add a SBF program to the genesis configuration with upgrades disabled. If the ledger already exists then this parameter is silently ignored. First argument can be a pubkey string or path to a keypair",
```

ananas-block:
the command doesn't work for me with the following input:
```
light-protocol$ ./cli/test_bin/run test-validator --sbf-program 2tzfijPBGbrR5PboyFUFKzfEoLTwdDSHUjANCw929wyt e2e_test
bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)

⠋ Performing setup tasks...

Starting test validator...
Error: program file does not exist: e2e_test
solana-test-validator exited with code 1
/home/ananas/dev/light-protocol/cli/dist/utils/process.js:104
                reject(new Error(`${commandBase} exited with code ${code}`));
                       ^

Error: solana-test-validator exited with code 1
    at ChildProcess.<anonymous> (/home/ananas/dev/light-protocol/cli/dist/utils/process.js:104:24)
    at ChildProcess.emit (node:events:514:28)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5)
```

can you add tests?

additionally can you adapt the template we use in `light init`  and test it with that as well?
(I think we don't have test coverage for light init right now at all which is not ideal. The inited project can either be in gitignore (thats how we did it for psp init commands) or the target folder.)

Does this work with multiple programs as inputs like this?
`light test-validator --sbf-program 2tzfijPBGbrR5PboyFUFKzfEoLTwdDSHUjANCw929wyt e2e_test --sbf-program asft4qfijPBGbrR5PboyFUFKzfEoLTwdDSHUjANCw929wyt other_test_program`

ananas-block:
Can you add a check for duplicates and collisions with our system programs?
(solana test-validator just doesn't deploy sth if it is a duplicate which is an annoying foot gun.)

sergeytimoshin:
> can you add tests?

Done (and while adding tests found a bug with the `indexer-port` flag: it only worked if we simulateonusely pass the `database-url` flag, fixed).

> additionally can you adapt the template we use in `light init` and test it with that as well? (I think we don't have test coverage for light init right now at all which is not ideal. The inited project can either be in gitignore (thats how we did it for psp init commands) or the target folder.)

Hm, the only place where we load sbf programs in template is https://github.com/Lightprotocol/compressed-program-template/blob/main/programs/%7B%7Brust-name%7D%7D/tests/test.rs#L27, but it's solana runtime, unrelated to our typescript cli, am I missed something?

> Does this work with multiple programs as inputs like this? `light test-validator --sbf-program 2tzfijPBGbrR5PboyFUFKzfEoLTwdDSHUjANCw929wyt e2e_test --sbf-program asft4qfijPBGbrR5PboyFUFKzfEoLTwdDSHUjANCw929wyt other_test_program`

Yes.

ananas-block:
Wrt template, you are right the template doesn't have a typescript right now. I thought it would  be nice to have the correct command to run the test command with a correctly configured light cli already even though its not used yet. But we can add this as well once we have added a ts test to the template.

cli/test/commands/test-validator/index.test.ts
```diff
@@ -0,0 +1,347 @@
...
+
+      const { stdout } = await exec(command);
+      expect(stdout).to.contain("Setup tasks completed successfully");
+
+      await new Promise((resolve) => setTimeout(resolve, 5000));
+
+      // Verify validator is running
+      const connection = new Connection(`http://localhost:${defaultRpcPort}`);
+      const version = await connection.getVersion();
+      expect(version).to.have.property("solana-core");
```
ananas-block, line 342: Let's add a check that the program was deployed correctly like:
```
let programAccountInfo = connection.getAccountInfo(testKeypair.publicKey);
assert(programAccountInfo);
assert(programAccountInfo!.isExecutable);
```

ananas-block (commented):
Thanks for adding the tests!
Looks great! Just have one more assert I would like to see.

cli/src/commands/test-validator/index.ts
```diff
@@ -15,7 +19,10 @@ class SetupCommand extends Command {
     '$ light test-validator --validator-args "--limit-ledger-size 50000000"',
```
ananas-block, line 19: can we add an example with `--sbf-program` as well?

ananas-block (approved):
lgtm feel free to merge once green!

PR #1394, opened 2024-12-15
chore: JS release t22

SwenSchaeferjohann:
- allow mintSize to be passed to createMint instruction. this is required if users want to createMint with t22 extensions
- release js 0.17.0

### DIFF
--- a/js/compressed-token/src/program.ts
+++ b/js/compressed-token/src/program.ts
@@ -209,6 +209,10 @@ export type CreateMintParams = {
      * Optional: The token program ID. Default: SPL Token Program ID
      */
     tokenProgramId?: PublicKey;
+    /**
+     * Optional: Mint size to use, defaults to MINT_SIZE
+     */
+    mintSize?: number;
 };
 
 /**
@@ -577,8 +581,16 @@ export class CompressedTokenProgram {
     static async createMint(
         params: CreateMintParams,
     ): Promise<TransactionInstruction[]> {
-        const { mint, authority, feePayer, rentExemptBalance, tokenProgramId } =
-            params;
+        const {
+            mint,
+            authority,
+            feePayer,
+            rentExemptBalance,
+            tokenProgramId,
+            freezeAuthority,
+            mintSize,
+        } = params;
+
         const tokenProgram = tokenProgramId ?? TOKEN_PROGRAM_ID;
 
         /// Create and initialize SPL Mint account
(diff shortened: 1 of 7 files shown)


js/compressed-token/src/program.ts
```diff
@@ -577,8 +581,16 @@ export class CompressedTokenProgram {
...
-        const { mint, authority, feePayer, rentExemptBalance, tokenProgramId } =
-            params;
+        const {
+            mint,
+            authority,
+            feePayer,
+            rentExemptBalance,
+            tokenProgramId,
+            freezeAuthority,
+            mintSize,
```
ananas-block, line 591: adding mint size like this leads to a weird devx.
To initialize token22 mints with extension the instruction order has to be `[createAccount, ...initExtensions, initMint, createTokenPoolIx]`. the current order is `[createMintAccountInstruction, initializeMintInstruction, createTokenPoolIx]` -> one would need to insert init token extension instructions between index 0 and 1 to use the function output to initialize a mint with token22 extensions.

ananas-block: We can do it in a different pr but ideally we should pass in the extension types and calculate space and create instructions inside of this function. In the mean time we could adjust the doc string so that it is possible to know where to inject create token 22 init extension instructions.
```
 /**
     * Construct createMint instructions for compressed tokens [createMintAccount, initializeMint, createTokenPool]
     */
 ```

SwenSchaeferjohann: added to docstring for now

ananas-block (approved):

ananas-block:
in your node token 2022 with metadata example the `createMint` function isn't used [see instructions](https://github.com/Lightprotocol/example-nodejs-client/blob/4172c5725388c4a2377ed7222c3c56cfe41427a0/src/scripts/mint-spl-22.ts#L59) .
If we don't use this function in this way can't we just remove the MINT_SIZE input?
(Seems unnecessary overhead to me and ripe for confusion.)

SwenSchaeferjohann:
> in your node token 2022 with metadata example the `createMint` function isn't used [see instructions](https://github.com/Lightprotocol/example-nodejs-client/blob/4172c5725388c4a2377ed7222c3c56cfe41427a0/src/scripts/mint-spl-22.ts#L59) . If we don't use this function in this way can't we just remove the MINT_SIZE input? (Seems unnecessary overhead to me and ripe for confusion.)

imo developers should be able to use either function- if we don't have the optional mintSize input, they can't use createMint. don't think it adds overhead/confusion since default use remains unchanged

good point regarding no example for it. added example for creating t22 with createMint [here](https://github.com/Lightprotocol/example-nodejs-client/blob/main/src/scripts/mint-spl-22-with-create-mint.ts)

PR #1407, opened 2024-12-19
feat: multiple `tokenPoolPda` accounts

ananas-block:
### Issue:
- the compress token program allows only one `tokenPoolPda` account to be created per token mint
- this causes a single shared write-lock for all tokens of a mint that get compressed or decompressed

### Solution:
- enable multiple `tokenPoolPda` accounts per mint
- don't allow too many accounts so that users can still decompress from all token pool accounts in one instruction if required
- as of this pr, the maximum of token pool accounts per mint is 5

### Changes:
- add instruction `add_token_pool`
- adapt `spl_token_pool_derivation` with a bump to derive multiple accounts
- for spl token compression loop over possible derivations until we find the correct one
- for spl token decompression, if the balance of the first `tokenPoolPda` is insufficient to support the decompression amount, loop over remaining accounts starting from index 0 and decompress. Loop until the complete decompression amount has been decompressed.

### DIFF
--- a/programs/compressed-token/src/burn.rs
+++ b/programs/compressed-token/src/burn.rs
@@ -87,22 +87,16 @@ pub fn burn_spl_from_pool_pda<'info>(
         }
         let mut remove_index = 0;
         for (index, i) in token_pool_bumps.iter().enumerate() {
-            match check_spl_token_pool_derivation(
-                mint_bytes.as_slice(),
-                &token_pool_pda.key(),
-                &[*i],
-            ) {
-                true => {
-                    burn(
-                        ctx,
-                        token_pool_pda.to_account_info(),
-                        withdrawal_amount,
-                        token_pool_amount,
-                    )?;
+            if check_spl_token_pool_derivation(mint_bytes.as_slice(), &token_pool_pda.key(), &[*i])
+            {
+                burn(
+                    ctx,
+                    token_pool_pda.to_account_info(),
+                    withdrawal_amount,
+                    token_pool_amount,
+                )?;
 
-                    remove_index = index;
-                }
-                false => {}
+                remove_index = index;
             }
         }
         token_pool_bumps.remove(remove_index);
--- a/programs/compressed-token/src/burn.rs
+++ b/programs/compressed-token/src/burn.rs
@@ -8,14 +8,14 @@ use light_system_program::{
 use light_utils::hash_to_bn254_field_size_be;
 
 use crate::{
-    constants::{NOT_FROZEN, NUM_MAX_POOL_ACCOUNTS},
+    constants::NOT_FROZEN,
     process_transfer::{
         add_token_data_to_input_compressed_accounts, cpi_execute_compressed_transaction_transfer,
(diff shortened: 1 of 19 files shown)


programs/compressed-token/src/spl_compression.rs
```diff
@@ -19,80 +20,131 @@ pub fn process_compression_or_decompression(
...
-        ctx.accounts
-            .token_program
-            .as_ref()
-            .unwrap()
-            .to_account_info(),
-        amount,
-    )
+    let mut token_pool_bumps = (0..NUM_MAX_POOL_ACCOUNTS).collect::<Vec<u8>>();
+
+    for i in 0..NUM_MAX_POOL_ACCOUNTS {
```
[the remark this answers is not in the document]
ananas-block: ty fixed

sergeytimoshin (approved):

programs/compressed-token/src/burn.rs
```diff
@@ -206,14 +221,20 @@ pub mod sdk {
         pub burn_amount: u64,
         pub signer_is_delegate: bool,
         pub is_token_22: bool,
+        pub token_pool_bump: u8,
+        pub additonal_pool_accounts: Vec<Pubkey>,
```
SwenSchaeferjohann, line 225: typo, should be `additional_pool_accounts`

programs/compressed-token/src/process_transfer.rs
```diff
@@ -777,11 +780,13 @@ pub mod transfer_sdk {
         compress_or_decompress_amount: Option<u64>,
         delegate_change_account_index: Option<u8>,
         lamports_change_account_merkle_tree: Option<Pubkey>,
+        accounts: &[Pubkey],
     ) -> (
         HashMap<Pubkey, usize>,
         CompressedTokenInstructionDataTransfer,
     ) {
         let mut additonal_accounts = Vec::new();
+        additonal_accounts.extend_from_slice(accounts);
```
SwenSchaeferjohann, line 789: same, typo

programs/compressed-token/src/instructions/burn.rs
```diff
@@ -24,9 +24,9 @@ pub struct BurnInstruction<'info> {
     /// CHECK: is used to burn tokens.
     #[account(mut)]
     pub mint: InterfaceAccount<'info, Mint>,
-    /// CHECK: (seed constraint) is derived from mint account.
-    #[account(mut, seeds = [POOL_SEED, mint.key().as_ref()], bump)]
-    pub token_pool_pda: InterfaceAccount<'info, TokenAccount>,
+    /// CHECK: in burn_spl_from_pool_pda.
```
robre, line 27: a check is not really being done!

ananas-block: fixed in [4f822f17e507d9f289432f161e535eaa9b8cd2b2](https://github.com/Lightprotocol/light-protocol/commit/4f822f17e507d9f289432f161e535eaa9b8cd2b2)

PR #1411, opened 2024-12-20
refactor: test indexer 

sergeytimoshin:


light-program-test/src/indexer/test_indexer.rs
```diff
@@ -0,0 +1,1272 @@
...
+        rpc: &mut R,
+    ) -> ProofRpcResult {
+        if compressed_accounts.is_some()
+            && ![1usize, 2usize, 3usize, 4usize, 8usize]
+                .contains(&compressed_accounts.unwrap().len())
+        {
+            panic!(
+                "compressed_accounts must be of length 1, 2, 3, 4 or 8 != {}",
+                compressed_accounts.unwrap().len()
+            )
```
ananas-block, line 310: we should remove the panics (we have these in multiple places including test indexer and test utils because I was lazy).

ananas-block: It's probably a good time to do it in this pr since we are defining a new trait hence let's make the trait return results instead of having functions panic.

merkle-tree/batched/src/merkle_tree.rs
```diff
@@ -138,7 +139,7 @@ impl BatchedMerkleTreeAccount {
             Pubkey::default(),
             rollover_fee,
         );
-        // inited address tree contains two elements.
+        // initialized address tree contains two elements.
```
ananas-block, line 142:
```suggestion
        // Initialized address tree contains two elements.
```

PR #1453, opened 2025-01-05
refactor: forester: batch ops

sergeytimoshin:


forester/src/batch_processor/common.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        match self.tree_type {
+            TreeType::BatchedAddress => {
+                info!("Processing address batch");
+                address::process_batch(&self.context).await
+            }
+            TreeType::BatchedState => {
+                info!("Processing state batch");
+                state::process_batch(&self.context).await
+            }
```
ananas-block, line 50: this needs to distinguish between an input(nullifier) queue batch being ready and an output queue batch being ready. The current implementation assumes that both are always ready at the same time which will fail in many cases.

forester/src/batch_processor/common.rs
```diff
@@ -0,0 +1,91 @@
...
+            _ => Err(error::BatchProcessError::UnsupportedTreeType(
+                self.tree_type,
+            )),
+        }
+    }
+
+    async fn verify_batch_ready(&self) -> bool {
```
ananas-block, line 57: doesn't cover the output queue whichs batches are independent of input queue batches.

The input queue is part of the Merkle tree account since we usually need access to Merkle tree roots to prove inclusion (except if the account hash wasn't inserted into the tree yet then we need the output queue to proof inclusion by index in the value array).

The output queue is a separate account with separate batches. You can check whether an output queue batch is ready with the same logic used for trees the data structure is similar.

ananas-block: The address Merkle tree doesn't have an output queue hence the current implementation conceptually works for address trees as is.

sergeytimoshin: Fixed, thanks!

forester/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "build": "cargo build",
     "test": "RUSTFLAGS=\"--cfg tokio_unstable -D warnings\" cargo test --package forester -- --test-threads=1 --nocapture",
-    "test-state-batch": "cargo test --package forester test_batched -- --test-threads=1 --nocapture",
+    "test-state-batched": "cargo test --package forester test_state_batched -- --test-threads=1 --nocapture",
+    "test-address-batched": "cargo test --package forester test_address_batched -- --test-threads=1 --nocapture",
```
ananas-block, line 9: could remove `--test-threads=1` since we use `#[serial]`

forester/src/batch_processor/common.rs
```diff
@@ -0,0 +1,169 @@
...
+            } else {
+                BatchReadyState::NotReady
+            };
+        }
+
+        if self.verify_input_queue_batch_ready(&mut rpc).await {
+            BatchReadyState::ReadyForNullify
+        } else if self.verify_output_queue_batch_ready(&mut rpc).await {
+            BatchReadyState::ReadyForAppend
+        } else {
```
ananas-block, line 111: Two thoughts:
1. what about switching this checking `verify_output_queue_batch_ready` with priority since I expect more output state creation than input state creation?
2. There could be a case where the forester isn't fast enough and one queue is filled so fast that the forester neglects the other queue. Probably not the right time to focus on this right now but makes sense to keep this in mind.

forester/tests/batched_state_test.rs
```diff
@@ -233,10 +274,83 @@ async fn test_batched() {
...
+        );
+
+        assert!(
+            merkle_tree
+                .get_metadata()
+                .queue_metadata
+                .next_full_batch_index
+                > 0,
+            "No batches were processed"
+        );
```
ananas-block, line 334:
```suggestion
        assert_eq!(
            merkle_tree
                .get_metadata()
                .queue_metadata
                .next_full_batch_index
                ,1,
            "No batches were processed"
        );
```

forester/tests/batched_state_test.rs
```diff
@@ -233,10 +274,83 @@ async fn test_batched() {
...
+
+        assert_ne!(
+            pre_root,
+            merkle_tree.get_root().unwrap(),
+            "Root should have changed"
+        );
+        assert!(
+            merkle_tree.root_history.len() > 1,
+            "Root history should contain multiple roots"
+        );
```
ananas-block, line 352: would be better to use exact numbers so that we can detect if the forester starts to behave differently.

ananas-block: Also we don't assert the root of the test indexer just that the root is not equal.
(The same for address tree asserts.)

ananas-block (approved):

PR #1468, opened 2025-01-08
refactor: test indexer

sergeytimoshin:

PR #1473, opened 2025-01-11
chore: add rust and program tests to test.sh

ananas-block:


sergeytimoshin (changes requested):
I'll prefer not to mix nx approach and cargo calls because we already have several tests executed by nx, so you will be running the same tests twice. I updated package.json to fully rely on nx in another branch:

https://github.com/Lightprotocol/light-protocol/pull/1468/commits/09ceb13d902de68a646fbc67dd2a237e0918e067#diff-4b116bbf0c0abde7e03e90966531b8412ef2892b44d99f071c4916a6772cf28cR9

PR #1474, opened 2025-01-12
chore: add xtask create state Merkle tree

ananas-block:


xtask/src/create_state_tree.rs
```diff
@@ -0,0 +1,108 @@
...
+    } else {
+        String::from("./target/state-trees/")
+    };
+
+    let mut mt_keypairs: Vec<Keypair> = vec![];
+    let mut nfq_keypairs: Vec<Keypair> = vec![];
+    let mut cpi_keypairs: Vec<Keypair> = vec![];
+
+    println!(
+        "ormat!( path, options.mt_pubkey) {:?}",
```
SwenSchaeferjohann, line 53:
```suggestion
        "format!( path, options.mt_pubkey) {:?}",
```

ananas-block:
```suggestion
        "path mt_pubkey {:?}",
```

xtask/src/create_state_tree.rs
```diff
@@ -0,0 +1,108 @@
...
+    println!("read cpi: {:?}", cpi_keypair.pubkey());
+    mt_keypairs.push(mt_keypair);
+    nfq_keypairs.push(nfq_keypair);
+    cpi_keypairs.push(cpi_keypair);
+
+    let payer = if let Some(payer) = options.payer {
+        read_keypair_file(&payer).unwrap_or_else(|_| panic!("{}{}.json", path, options.cpi_pubkey))
+    } else {
+        read_keypair_file("/home/ananas/.config/solana/id.json")
+            .expect("Default payer keypair not found in /home/ananas/.config/solana/id.json")
```
SwenSchaeferjohann, line 71: best to replace with generic path

- eg `~/.config/solana/id.json`
- or using "dirs" crate
 ```
use dirs::home_dir;

// Construct the path to the keypair file in the user's home directory
let keypair_path: PathBuf = home_dir()
   .expect("Could not find home directory")
   .join(".config/solana/id.json");
```
- or with an env var `SOLANA_KEYPAIR_PATH`

SwenSchaeferjohann (commented):
cool, just one comment below

xtask/src/create_state_tree.rs
ananas-block: fixed with home_dir

SwenSchaeferjohann (approved):

PR #1486, opened 2025-01-15
chore: abbreviate batched trees from methods

ananas-block:
Changes:
1. rename `state_tree_from_bytes_mut`  -> `state_from_bytes`
2. rename `address_tree_from_bytes_mut` -> `address_from_bytes`
3. rename `output_queue_from_bytes_mut` -> `output_from_bytes`
4. rename `from_bytes_mut` -> `from_bytes`
   (we are not going to add a non mutable from bytes version) 
5. add check and error whether Merkle tree is full
    1. add `tree_capacity` to  `QueueAccount` to check tree fullness
    2. add `capacity` to `MerkleTreeAccount` to check fullness at address queue insertion


program-libs/batched-merkle-tree/src/queue.rs
```diff
@@ -49,6 +46,8 @@ pub struct BatchedQueueMetadata {
     /// next_index in queue is ahead or equal to next index in the associated
     /// batched Merkle tree account.
     pub next_index: u64,
+    /// Height of the associated tree.
```
sergeytimoshin, line 49:
```suggestion
    /// Maximum number of leaves that can fit in the tree, calculated as 2^height.
    /// For example, a tree with height 3 can hold up to 8 leaves.
```

sergeytimoshin (approved):

PR #1492, opened 2025-01-19
refactor: batched trees to 2 batches

ananas-block:
Issue:
- Merkle trees `from_account_info` took `1.5k` CU
- rust ci tests were not executing `test_e2e` and ` test_simulate_transactions`

Solution:
- reduce the number of Ref casts -> reduces `from_account_info` calls to `800-900` CU 
     1. bloom filter stores  are just bytes slices no need to wrap these into a zero copy slice) 
     2. make two `Batch` structs part of `BatchMetadata` -> removes 2 casts
     3. remove unnecessary account size check in `from_bytes` (saves 200 CU)


program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -833,13 +835,13 @@ impl<'a> BatchedMerkleTreeAccount<'a> {
...
         value: &[u8; 32],
     ) -> Result<(), BatchedMerkleTreeError> {
-        let num_bloom_filters = self.bloom_filter_stores.len();
-        for i in 0..num_bloom_filters {
-            let bloom_filter_store = self.bloom_filter_stores[i].as_mut_slice();
-            let batch = &self.batches[i];
-            if !batch.bloom_filter_is_zeroed() {
-                batch.check_non_inclusion(value, bloom_filter_store)?;
-            }
+        for i in 0..2 {
```
sergeytimoshin, line 838: Can we replace 0..2 with num_batches?

sergeytimoshin (approved):

program-libs/batched-merkle-tree/src/merkle_tree.rs
ananas-block:
```suggestion
        for i in 0..self.queue_metadata.batches.len() {
```

PR #1514, opened 2025-01-26
refactor: queue index to bool

ananas-block:
Issue:
- `Option<QueueIndex>` was a place holder variable in anticipation to prove inclusion by index
- Now we do prove inclusion by index but a simple boolean is sufficient there is no need for the struct `QueueIndex`.
- `Option<QueueIndex>` is required to be set as Some() to signal prove by index but contents of the `QueueIndex` struct are not used which is confusing
- options are a pain and expensive in zero copy deserialization

Current State:
- on mainnet Option<QueueIndex> is enforced as None

Solution:
- replace `Option<QueueIndex>` with `bool`
- no breaking changes in instruction data borsh serialization since on mainnet we enforce `queue_index` to be `None` which is serialized in 1 byte as 0 the same as a bool = `false`
- this will make  instruction data  (de)serialization cheaper and easier


SwenSchaeferjohann (approved):
lgtm

PR #1525, opened 2025-01-27
refactor: move PhotonIndexer to light-client crate

sergeytimoshin:


sdk-libs/client/src/indexer/photon_indexer.rs
```diff
@@ -252,3 +246,10 @@ impl<R: RpcConnection> Indexer<R> for PhotonIndexer<R> {
         todo!()
     }
 }
+
+fn decode_hash(account: &str) -> [u8; 32] {
+    let bytes = bs58::decode(account).into_vec().unwrap();
+    let mut arr = [0u8; 32];
+    arr.copy_from_slice(&bytes);
+    arr
+}
```
ananas-block, line 255: The following avoids the vector allocation. (Just a suggestion below didn't test it.)
```suggestion
fn decode_hash(account: &str) -> [u8; 32] {
    let mut arr = [0u8; 32];
    bs58::decode(account)
        .into(&mut arr)
        .expect("Failed to decode base58 string");
    arr
}
```

ananas-block: would be better to return a result

ananas-block (commented):
Just the one comment else lgtm

sdk-libs/client/src/indexer/photon_indexer.rs
sergeytimoshin: Thanks, done https://github.com/Lightprotocol/light-protocol/pull/1525/commits/301544dd7b95e06a376e4471e723e0313d3fba2a.

We already have [cleaner implementation for conversions like this](https://github.com/Lightprotocol/light-protocol/blob/main/sdk-libs/client/src/photon_rpc/types.rs#L25-L37) btw, but there is a mess with photon_client & sdk types that needs to be untangled first. 
I started doing it here, but its not finished yet https://github.com/Lightprotocol/light-protocol/pull/1515

ananas-block (approved):

PR #1526, opened 2025-01-27
feat: rate limiter

sergeytimoshin:
Implemented rate limiter for rpc calls.
Forester can be run with 3 optional flags:
```
--rpc-rate-limit <RPC_RATE_LIMIT>
--photon-rate-limit <PHOTON_RATE_LIMIT>
--send-tx-rate-limit <SEND_TX_RATE_LIMIT>
```

### DIFF
--- a/sdk-libs/client/Cargo.toml
+++ b/sdk-libs/client/Cargo.toml
@@ -34,6 +34,9 @@ num-bigint = { workspace = true }
 num-traits = { workspace = true }
 reqwest = { workspace = true }
 
+governor = "0.8.0"
+
+
 [dev-dependencies]
 light-test-utils = { workspace = true, features=["devenv"]}
 light-program-test = { workspace = true }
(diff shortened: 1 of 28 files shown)


sdk-libs/client/Cargo.toml
```diff
@@ -34,11 +34,14 @@ num-bigint = { workspace = true }
 num-traits = { workspace = true }
 reqwest = { workspace = true }
 
+governor = "0.8.0"
```
ananas-block, line 37: should we make this a workspace dep?

ananas-block (approved):
One nit. Approved but didn't merge since I assume that you want to merge 1525 before this one.

sdk-libs/client/Cargo.toml
sergeytimoshin: It's only used in one crate and shouldn't be used anywhere else, so I think we're fine here.

PR #1531, opened 2025-01-30
refactor: restore name-service-without-macros example

sergeytimoshin:
1. Refactored `get_compressed_accounts_by_owner` method in `test_indexer.rs` and `photon_indexer.rs` to return structured data instead of raw hashes.
2. Adjusted test logic in name-service-without-macros example.


sergeytimoshin:
Closing this, continue in https://github.com/Lightprotocol/light-protocol/pull/1534

PR #1563, opened 2025-02-07
feat: add batch compress tokens

ananas-block:
## State
- ready, just in draft until final review to allow a clean commit for the audit
- kept as a backup branch this is the latest pr https://github.com/Lightprotocol/light-protocol/pull/1718

### DIFF
--- a/program-tests/compressed-token-test/tests/test.rs
+++ b/program-tests/compressed-token-test/tests/test.rs
@@ -1204,7 +1204,6 @@ async fn test_mint_to_failing() {
             assert!(result
                 .to_string()
                 .contains("Error processing Instruction 0: Cross-program invocation with unauthorized signer or writable account"));
-            // assert_rpc_error(result, 0, 0).unwrap();
         }
         // 9. Invalid Merkle tree.
         {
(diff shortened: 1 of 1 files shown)


programs/compressed-token/src/lib.rs
```diff
@@ -79,7 +81,30 @@ pub mod light_compressed_token {
...
+        ctx: Context<'_, '_, '_, 'info, MintToInstruction<'info>>,
+        inputs: Vec<u8>,
+    ) -> Result<()> {
+        let (inputs, _) = batch_compress::BatchCompressInstructionData::zero_copy_at(&inputs)
+            .map_err(ProgramError::from)?;
+
+        process_mint_to::<COMPRESS>(
+            ctx,
+            inputs.pubkeys.as_slice(),
+            inputs.amounts.as_slice(),
```
SwenSchaeferjohann, line 104: @ananas-block  is `inputs.amounts` having multiple amounts? 
ideally for this ix we just have a single `inputs.amount` and apply it for all recipients

ananas-block: yeah I can add an option for that

PR #1577, opened 2025-02-10
chore: rename light-utils -> light-compressed-account, move account checks to light-account-checks

ananas-block:


program-libs/account-checks/src/discriminator.rs
```diff
@@ -0,0 +1,10 @@
+pub const ANCHOR_DISCRIMINATOR_LEN: usize = 8;
+
+// pub trait AnchorDiscriminator = Discriminator<[u8; 8]>;
```
SwenSchaeferjohann, line 3: leftover?

ananas-block: Yeah

ananas-block:
```suggestion
```

PR #1588, opened 2025-02-15
fix: precision in js rpc client

SwenSchaeferjohann:
Issue:
	•	The Node.js Fetch API’s `response.json()` internally uses `JSON.parse()`, which converts values larger than `Number.MAX_SAFE_INTEGER` to `Number`, causing precision loss.

Fix:
	•	Replace `response.json()` with `json-bigint.parse()`.
	•	Explicitly coerce values in `BNFromStringOrNumberOrBigNumber`.
	
	
Other changes:
	•	add `logger.ts`
	•	add test to assert correct coercion


js/stateless.js/src/rpc-interface.ts
```diff
@@ -200,16 +201,69 @@ const BN254FromString = coerce(instance(BN), string(), value => {
...
-    // Check if the number is safe
-    if (Number.isSafeInteger(value)) {
-        return bn(value);
-    } else {
-        // Convert to string if the number is unsafe
-        return bn(value.toString(), 10);
+/**
+ * @internal
+ */
+function convertBigNumberObjectToBN(obj: any) {
```
ananas-block, line 207: can't we use a library for this?

js/stateless.js/tests/e2e/safe-conversion.test.ts
```diff
@@ -0,0 +1,151 @@
...
+        if ('error' in res) {
+            throw new Error('error in res');
+        }
+        if (res.result.value === null) {
+            throw new Error('not implemented: NULL result');
+        }
+
+        const tokenAmount = res.result.value.items[0].tokenData.amount;
+
+        expect(tokenAmount.toString()).toBe('9999999820999994000');
```
ananas-block, line 79: we should do a test with u64::MAX since this is the max possible value

js/stateless.js/src/rpc-interface.ts
```diff
@@ -200,16 +201,69 @@ const BN254FromString = coerce(instance(BN), string(), value => {
...
+            if (!Number.isSafeInteger(value)) {
+                throw new Error(`Unsafe integer. Precision loss: ${value}`);
+            }
+            return new BN(value); // Safe number → BN
+        }
+        if (typeof value === 'object') {
+            if (value instanceof BN) {
+                return value;
+            } else if ('s' in value && 'e' in value && 'c' in value) {
+                return convertBigNumberObjectToBN(value);
```
ananas-block, line 259: chat gpt recommends to use bignumberjs instead sth like:

```
const BigNumber = require('bignumber.js'); 
let bn = new BigNumber(value)
return new BN(bn.toString(), 10)
```

js/stateless.js/src/rpc-interface.ts
SwenSchaeferjohann: simplified by doing in preprocessing what otherwise solana's json rpc api does server-side: wrapping bignumbers into quotes. this is simple, and most lightweight

js/stateless.js/src/rpc.ts
```diff
@@ -323,11 +337,15 @@ export const rpcRequest = async (
         throw new Error(`HTTP error! status: ${response.status}`);
     }
 
+    const text = await response.text();
+
+    const wrappedJsonString = wrapBigNumbersAsStrings(text);
```
ananas-block, line 342: ideally we shouldn't need helpers like this it feels hacky.
Did you check how Solana ts rpc handles big numbers in rpc requests?

SwenSchaeferjohann: solana does this server side - added more tests, there's plenty. imo we're good

js/stateless.js/src/logger.ts
```diff
@@ -0,0 +1,19 @@
...
+// // call logger.info('text', { text: 'test' })
+// export const logger = winston.createLogger({
+//     level: 'info',
+//     format: winston.format.combine(
+//         winston.format.timestamp(),
+//         winston.format.prettyPrint(),
+//     ),
+//     transports: [
+//         new winston.transports.File({ filename: 'combined.log' }),
+//         new winston.transports.Console({
```
ananas-block, line 15: leftover?

ananas-block (approved):

PR #1603, opened 2025-02-22
chore: cfg anchorless compressed-account sdk

SwenSchaeferjohann:


SwenSchaeferjohann:
the methods don't account for lamports yet

PR #1604, opened 2025-02-23
feat: add next nullifier index

ananas-block:
Issue:
- when indexing the nullifier event it was difficult to remove the correct leaves from the nullifier queue

Changes:
- add `nullifier_next_index` to simplify removing values from the input queue by using it as a continuous index
- this changes the BatchedMerkleTreeStruct -> regenerated accounts


program-libs/batched-merkle-tree/src/merkle_tree_metadata.rs
```diff
@@ -162,6 +164,7 @@ impl BatchedMerkleTreeMetadata {
             hashed_pubkey: hash_to_bn254_field_size_be(&tree_pubkey.to_bytes())
                 .unwrap()
                 .0,
+            nullifier_next_index: 0,
```
SwenSchaeferjohann, line 167: if TreeType::BatchedAddress does this need to be 0 or 2?

ananas-block: it's always 0. The queue is always empty at the start for both address and state. Additionally the nullifier next index is not relevant for the address Merkle tree since we append elements and only have 1 queue.

SwenSchaeferjohann (approved):

PR #1614, opened 2025-02-26
fix: parse event from transaction

ananas-block:
General:
- fix parsing light event so that it can deserialize multiple events within one outer instruction
- added randomized unit tests
- functional e2e tests  (including addresses, inputs, outputs)

Changes:
-  created `create_nullifier` function and am using it in light-batched-merkle-tree and light-compressed-account (currently we hash it manually)
- refactor `create_tx_hash` into `tx_hash` file
- refactor `event` file into `indexer_event` directory with separate errors


SwenSchaeferjohann (approved):
lgtm

PR #1631, opened 2025-03-05
refactor: forester(send_transaction): add timeout checks

sergeytimoshin:


forester/src/send_transaction.rs
```diff
@@ -211,10 +210,16 @@ pub async fn send_batched_transactions<T: TransactionBuilder, R: RpcConnection>(
         })
         .collect();
 
+    let buffer_duration = Duration::from_secs(2);
+    let adjusted_timeout = if config.retry_config.timeout > buffer_duration {
+        config.retry_config.timeout - buffer_duration
+    } else {
+        return Ok(0);
```
ananas-block, line 217: Does this mean if we don't set a timeout in the config we only retry once?

ananas-block (approved):
Just one comment whether I got it right and its intentional.

forester/src/send_transaction.rs
sergeytimoshin: In theory it is potential footgun, but timeout is non-optional and we set as a [duration of light slot](https://github.com/Lightprotocol/light-protocol/blob/sergey/forester-tx-timeout-check/forester/src/epoch_manager.rs#L851-L864), so I think we safe here.

PR #1633, opened 2025-03-06
feat: optimize forester by parallelizing zkp proof generation for batch append and nullify

sergeytimoshin:


forester-utils/src/instructions/state_batch_append.rs
```diff
@@ -0,0 +1,232 @@
...
+            .collect::<Vec<[u8; 32]>>();
+        let merkle_proofs = batch_elements
+            .iter()
+            .map(|x| x.proof.clone())
+            .collect::<Vec<Vec<[u8; 32]>>>();
+
+        let batch_root = batch_elements.first().unwrap().root;
+        let proof_future = generate_zkp_proof(
+            batch_root,
+            current_index as u32,
```
ananas-block, line 140: `current_index` has to be incremented by zkp batch size.

forester-utils/src/instructions/state_batch_append.rs
```diff
@@ -0,0 +1,232 @@
...
+            .map(|x| x.account_hash)
+            .collect::<Vec<[u8; 32]>>();
+        let merkle_proofs = batch_elements
+            .iter()
+            .map(|x| x.proof.clone())
+            .collect::<Vec<Vec<[u8; 32]>>>();
+
+        let batch_root = batch_elements.first().unwrap().root;
+        let proof_future = generate_zkp_proof(
+            batch_root,
```
ananas-block, line 139: has to be the new root of the previous prover request

forester-utils/src/instructions/state_batch_append.rs
```diff
@@ -0,0 +1,232 @@
...
+            current_index as u32,
+            leaves,
+            *leaves_hash_chain,
+            old_leaves,
+            merkle_proofs,
```
ananas-block, line 144: Merkle proofs need to be patched with the change logs of all previous iterations, since the indexer is on current onchain state.
-> `generate_zkp_proof` needs to return changelogs or just mutate a changelogs vector that is shared between proof requests.

forester-utils/src/account_zero_copy.rs
```diff
@@ -0,0 +1,93 @@
...
+use light_hash_set::HashSet;
+use light_hasher::Hasher;
+use light_indexed_merkle_tree::copy::IndexedMerkleTreeCopy;
+use num_traits::{CheckedAdd, CheckedSub, ToBytes, Unsigned};
+use solana_sdk::account::Account;
+
+#[derive(Debug, Clone)]
+pub struct AccountZeroCopy<'a, T> {
+    pub account: Pin<Box<Account>>,
+    deserialized: *const T,
```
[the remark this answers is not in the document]
sergeytimoshin: This code is very old though and we still use it in program-tests/*, I just moved it to a separate file from the https://github.com/Lightprotocol/light-protocol/blob/main/forester-utils/src/lib.rs#L37-L117

If we want to remove it, let's do it in separate PR.

forester/src/batch_processor/state.rs
```diff
@@ -15,77 +18,188 @@ use crate::{
...
+        instruction_data_vec.len()
     );
 
-    rpc.create_and_send_transaction(
-        &[instruction],
-        &context.authority.pubkey(),
-        &[&context.authority],
-    )
-    .await?;
+    for (idx, instruction_data) in instruction_data_vec.iter().enumerate() {
```
ananas-block, line 56: we could iterate over instruction_data_vec in chunks eg 8 (I am not sure how many we can fit into one tx).
```suggestion
    for (idx, instruction_data) in instruction_data_vec.iter().chunks(8).enumerate() {
```

forester/src/batch_processor/state.rs
```diff
@@ -15,77 +18,188 @@ use crate::{
...
+        output_queue = %context.output_queue,
+    ), skip(context, rpc))
+]
 pub(crate) async fn perform_append<R: RpcConnection, I: Indexer<R> + IndexerType<R>>(
     context: &BatchContext<R, I>,
     rpc: &mut R,
 ) -> Result<()> {
-    debug!("perform_append");
-    let instruction_data = create_append_batch_ix_data(
+    let instruction_data_vec = create_append_batch_ix_data(
```
ananas-block, line 34: Nice that we compute proofs in parallel now.

Some considerations on edge cases and harmonizing parallel proof gen with sending transactions:
1. in the current state we compute proofs over all available queue data, in case that combined proof gen takes longer than say half of the forester slot we cannot send any transactions.
    Possible mitigation:
    1. compute only as many proofs as we can fit into one transaction (not enough for mainnet performance)
    2. stop creating new proof jobs once proof_start_time + average_proving_time > 15s (half a forester slot)
    3. Ideally we probably should continuously create new proof jobs until timeout (eg half slot time) and then send all transactions in sequence with as many proofs per tx as possible each.
       -> we should mix nullification and append jobs to avoid a queue being neglegted (both need to depend on the same shared changelogs since they use the same indexer state, changelogs become invalid as soon as the first tree update transaction is indexed)

ananas-block: > send 25 ordered tx in 15 seconds

Is probably not possible without either bundling tx which is expensive or an onchain cache similar to a changelog.

forester-utils/src/instructions/state_batch_append.rs
```diff
@@ -0,0 +1,232 @@
...
+    wait_for_indexer(rpc, indexer).await?;
+
+    let total_elements = zkp_batch_size as usize * leaves_hash_chains.len();
+    let offset = merkle_tree_next_index;
+
+    let queue_elements = indexer
+        .get_queue_elements(
+            merkle_tree_pubkey.to_bytes(),
+            QueueType::BatchedOutput,
+            total_elements as u16,
```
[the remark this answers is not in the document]
ananas-block: what do we do if the returned elements aren't all used in proofgen?

forester-utils/src/account_zero_copy.rs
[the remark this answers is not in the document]
ananas-block: albeit ugly the worst that can happen in tests is that we waste time debugging, forester-utils is used in production though and in prod this is dangerous.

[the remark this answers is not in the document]
ananas-block: actually it's fine I confused it with another similar abstraction which just casts bytes this one uses the correct deserialization methods.

forester-utils/src/instructions/state_batch_append.rs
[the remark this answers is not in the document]
sergeytimoshin: we trying to use all of them https://github.com/Lightprotocol/light-protocol/blob/sergey/async-proof-batch/forester-utils/src/instructions/state_batch_append.rs#L56-L63

ananas-block (approved):

PR #1634, opened 2025-03-06
chore: update photon api

ananas-block:


sdk-libs/client/src/indexer/photon_indexer.rs
```diff
@@ -782,6 +782,7 @@ impl<R: RpcConnection> Indexer<R> for PhotonIndexer<R> {
             .await?;
 
             let result = Self::extract_result("get_validity_proof", result.result)?;
+            println!("result {:?}", result);
```
sergeytimoshin, line 785: leftover?

sergeytimoshin (commented):
I think it makes sense to enable forester CI callback to  trigger on changes in the /sdk-libs/photon-api/*

sdk-libs/client/src/indexer/photon_indexer.rs
ananas-block:
```suggestion
```

PR #1635, opened 2025-03-08
refactor: remove rate limiter

sergeytimoshin:
Rate limiter in its current form hinders, not helps: its inclusion can lead to a significant decrease in transaction landing. 
To limit the number of requests we can use rpc pool, which indirectly solves the same problem.


ananas-block:
how does the rpc pool solve the rate limiting problem?

sergeytimoshin:
> how does the rpc pool solve the rate limiting problem?

Our rpc pool maintains a [fixed set of connections](https://github.com/Lightprotocol/light-protocol/blob/main/sdk-libs/client/src/rpc_pool.rs#L83). 
When all (max_size) connections are being used by concurrent threads, new requests must wait in queue until a connection becomes available again.

PR #1650, opened 2025-03-18
refactor: LightHasher and AsByteVec -> ToByteArray

ananas-block:
based on https://github.com/Lightprotocol/light-protocol/pull/1652

### General Idea

- Trait ToByteArray:
    1. ToByteArray -> [u8;32]
    2. ToByteArrays -> [[u8;32]; NUM_FIELDS]
    3. const NumFields -> usize (can be used to get ToByteArrays)
- Trait DataHasher Hash -> [u8;32]
- everything is BE hashed

**Derive macro LightHasher derives  ToByteArray and DataHasher.**

- Macro Attributes:
    1. hash
        hash to bn254 field size (less than 254 bit), hash with keccak256 and truncate to 31 bytes
    2. skip
        ignore field
    3. flatten
        flatten nested struct or vector

**Macro Rules**
- Struct:
  - every field must implement ToByteArray
  - impl ToByteArray for Struct -> returns hash of all fields
  - impl DataHasher for Struct -> returns hash of all fields
- Options (primitive types PT):
    - Option<PT> -> [u8;32] -> Some: [32 - type_bytes_len..] 32 - index type_bytes_len -1 = [1] (BE prefix) , None: [0;32]
- Option (General):
    - Option<T> T must implement Hash -> Some: Hash(T::hash), None: [0u8;32]
- Nested Structs:
    - to_byte_array -> hash of all fields
- Arrays (u8):
    1. LEN < 32 implementation of ToByteArray is provided
    2. LEN >= 32  needs to be handled (can be truncated or implement custom ToByteArray)
- Arrays:
    1. if elements implement ToByteArray and are less than 13, hash of all elements
    2. More elements than 13 -> manual implementation or hash to field size
    (Not implemented in derive macro but in `impl_to_byte_array_for_array`)
- Vec<T>:
    - we do not provide a blanket implementation since it could fail in production once a vector fills up
    - users need to hash to field size or do a manual implementation
- Strings:
    - we do not provide a blanket implementation since it could fail in production once a string fills up
    - users need to hash to field size or do a manual implementation
- Enums, References, SmartPointers:
    - Not supported

### Changes:
- `CompressedAccount`
    - `hash_to_bn254_field_size_be` : return `Option<[u8;32]>` -> return `[u8;32]` (cannot return None)
    - 
- `light-hasher`
    1. add HashToFieldSize trait and implement it for: `String`, `solana_program::Pubkey`, `Vec<T>`,
    2. refactor `ToByteVec` -> `ToByteArray`
        - Option<T> 
            - Some(PrimitiveType) prefixed with 1
            - Some(T) -> Poseidon::hash(T::to_byte_array())
            - None -> [0u8;32]
        - String
            - supported until byte length 31 -> return Err()
        -  
    4. 
- `LightHasher`
    - `#[truncate]` -> `#[hash]`
    -


ananas-block:
TODO:
- rename LightHasher -> PoseidonHasher
- consider adding a very cheap ShaHasher that just flattens everything

programs/account-compression/src/processor/insert_into_queues.rs
```diff
@@ -31,6 +31,8 @@ pub fn process_insert_into_queues<'a, 'b, 'c: 'info, 'info>(
     }
     let current_slot = Clock::get()?.slot;
     msg!("insert_nullifiers {:?}", inputs.nullifiers.len());
+    msg!("insert_leaves {:?}", inputs.leaves.len());
+    msg!("insert_addresses {:?}", inputs.addresses.len());
```
SwenSchaeferjohann, line 35: leftovers?

SwenSchaeferjohann (approved):

programs/account-compression/src/processor/insert_into_queues.rs
ananas-block: on purpose see https://github.com/Lightprotocol/light-protocol/pull/1652 comment

PR #1654, opened 2025-03-22
feat: TS add freeze and thaw

SwenSchaeferjohann:


js/compressed-token/src/actions/freeze.ts
```diff
@@ -0,0 +1,67 @@
...
+        recentInputStateRootIndices: proof.rootIndices,
+        recentValidityProof: proof.compressedProof,
+        mint,
+    });
+
+    const { blockhash } = await rpc.getLatestBlockhash();
+    const additionalSigners = dedupeSigner(payer, [freezeAuthority]);
+
+    const signedTx = buildAndSignTx(
+        [ComputeBudgetProgram.setComputeUnitLimit({ units: 1_000_000 }), ix],
```
ananas-block, line 60: would it make sense to make compute units variable?
(pass compute units as optional variable, 1mm is too many for sure.)

PR #1662, opened 2025-03-28
refactor: light-sdk, Light Account and LightAccountInfo

ananas-block:
### Changes:
- rename `RemainingAccounts` -> `CpiAccounts`
- rename `verify` ->  `verify_borsh` , pull in signer seeds and generate bump seed at compile time in `invoke_light_system_program`
- remove `verify_light_accounts` instead we can do `CAccount.to_account_info()` and use `verify_light_account_infos`
- rename LightAccountInfo -> `CAccountInfo`, split it into `CInAccountInfo` and `COutAccountInfo` for cheaper zero copy
    - removed options for lamports for cheaper zero copy
    - note that `CAccountInfo` will need to move into `light-compressed-account` once we add the pda optimized system program instruction
    - `SystemInfoInstructionData` is unused
-  split up `LightAccountMeta` into 4 specialized structs:
    1. InputAccountMeta (context, address, root_index, output_merkle_tree_index)
        (A pda with address, output, and no lamports is assumed to be the standard case.)
    2. InputAccountMetaNoLamportsNoAddress (context, root_index, output_merkle_tree_index)
    3. InputAccountMetaWithLamportsNoAddress (context, root_index, output_merkle_tree_index)
    4. InputAccountMetaWithLamports (context, lamports, address, root_index, output_merkle_tree_index)
    - `InputAccountMetaTrait` can be used onchain to instantiate a `CAccount` or `CAccountInfo`
- removed `LightAccountMeta` from `LightInstructionData` including it's deserialization function imo it should be part of the devs custom instruction data see tests for usage

### Notes:
- experimented with new naming because the `Light` prefix is makes names even longer, happy to chat about naming

### DIFF
--- a/program-tests/sdk-test/src/update_pda.rs
+++ b/program-tests/sdk-test/src/update_pda.rs
@@ -1,0 +1,67 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use light_sdk::{
+    account::CBorshAccount,
+    account_meta::InputAccountMeta,
+    error::LightSdkError,
+    instruction_data::LightInstructionData,
+    system_accounts::{LightCpiAccounts, SystemAccountInfoConfig},
+    verify::verify_light_account_infos,
+};
+use solana_program::account_info::AccountInfo;
+
+use crate::create_pda::MyCompressedAccount;
+
+/// CU usage:
+/// - sdk pre system program  10,902k CU
+/// - total with V2 tree: 78,074 CU (proof by index)
+pub fn update_pda<const BATCHED: bool>(
+    accounts: &[AccountInfo],
+    instruction_data: &[u8],
+) -> Result<(), LightSdkError> {
+    let mut instruction_data = instruction_data;
+    let instruction_data = UpdatePdaInstructionData::deserialize(&mut instruction_data)
+        .map_err(|_| LightSdkError::Borsh)?;
+
+    let program_id = crate::ID.into();
+    let mut my_compressed_account = CBorshAccount::<'_, MyCompressedAccount>::new_mut(
+        &program_id,
+        &instruction_data.my_compressed_account.meta,
+        MyCompressedAccount {
+            data: instruction_data.my_compressed_account.data,
+        },
+    )?;
+
+    my_compressed_account.data = instruction_data.new_data;
+
+    let config = SystemAccountInfoConfig {
+        self_program: crate::ID,
+        cpi_context: false,
+        sol_pool_pda: false,
+        sol_compression_recipient: false,
+    };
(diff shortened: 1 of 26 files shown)


sdk-libs/sdk/src/system_accounts.rs
```diff
@@ -38,10 +36,10 @@ impl<'c, 'info> LightCpiAccounts<'c, 'info> {
...
         program_id: Pubkey,
     ) -> Result<Self> {
-        if accounts.len() < SYSTEM_ACCOUNTS_LEN {
-            solana_program::msg!("accounts len {}", accounts.len());
-            return Err(LightSdkError::FewerAccountsThanSystemAccounts);
-        }
+        // if accounts.len() < SYSTEM_ACCOUNTS_LEN {
+        //     solana_program::msg!("accounts len {}", accounts.len());
+        //     return Err(LightSdkError::FewerAccountsThanSystemAccounts);
+        // }
```
ananas-block, line 42: left on purpose could be a good sanity check but don't work right now

sdk-libs/sdk/src/transfer.rs
```diff
@@ -9,28 +9,41 @@ use crate::{
...
+    // - If an account has no output compressed account have to transfer all lamports from it.
+    // - However the account does not have an output balance to measure the difference.
+    // - We could solve this by using the output balance anyway but skipping output values
+    //      which only use lamports and no data (program owned compressed must have data).
+    // else if let Some(input) = from.input.as_mut() {
+    //     input.lamports = input
+    //         .lamports
+    //         .checked_sub(lamports)
+    //         .ok_or(LightSdkError::TransferFromInsufficientLamports)?
+    // }
```
ananas-block, line 34: left for a future pr

program-tests/sdk-test/src/update_pda.rs
```diff
@@ -0,0 +1,67 @@
...
+    )?;
+
+    my_compressed_account.data = instruction_data.new_data;
+
+    let config = SystemAccountInfoConfig {
+        self_program: crate::ID,
+        cpi_context: false,
+        sol_pool_pda: false,
+        sol_compression_recipient: false,
+    };
```
SwenSchaeferjohann, line 41: can use `SystemAccountInfoConfig::new(crate::ID);`

SwenSchaeferjohann (commented):
Looks great overall! Collected all suggestions [here](https://www.notion.so/program-SDK-suggestions-1c42e76ac384806889aecf4cd7eecfc9?pvs=4), using the anchor-less test-program to showcase.

SwenSchaeferjohann (approved):

PR #1677, opened 2025-04-07
feat: update batched merkle tree with changelogs

sergeytimoshin:


program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
+        // First look for an existing entry to replace
+        for i in 0..self.changelog.len() {
+            let existing = self.changelog[i];
+            if existing.hash_chain_index == hash_chain_index && 
+                existing.pending_batch_index == pending_batch_index {
+                // Replace existing entry
+                self.changelog[i] = entry;
+                return;
+            }
+        }
```
ananas-block, line 151: this is not necessary, the cyclic vec takes care of overwriting existing values.

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
 
 impl Discriminator<ANCHOR_DISCRIMINATOR_LEN> for BatchedMerkleTreeAccount<'_> {
     const DISCRIMINATOR: [u8; 8] = *b"BatchMta";
 }
 
 impl<'a> BatchedMerkleTreeAccount<'a> {
+    // Common implementation that works for both test and non-test modes
+    
+    /// Checks if a changelog entry is applicable to the current state
+    pub fn is_changelog_entry_applicable(&self, entry: &crate::changelog::BatchChangelog) -> bool {
```
ananas-block, line 111: could it make sense to move all changelog related functions into a separate `changelog.rs` file to make the diff as clean as possible?

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
+            zkp_batch_size: self.metadata.queue_batches.zkp_batch_size,
+            old_next_index: self.metadata.next_index,
+            new_next_index: self.metadata.next_index,
+            root_index: self.root_history.last_index() as u32,
+            output_queue_pubkey: None,
+        }))
+    }
+    
+    /// Find all changelog entries that are applicable to the current state
+    pub fn find_applicable_changelog_entries(&self, current_root: &[u8; 32], current_seq: u64)
```
ananas-block, line 187: probably more efficient to `.filter`  instead of allocating a new vector

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
+        // Mark the entries as processed by zeroing them out
+        let zeroed_entry = crate::changelog::BatchChangelog {
+            old_root: [0u8; 32],
+            new_root: [0u8; 32],
+            leaves_hash_chain: [0u8; 32],
+            hash_chain_index: 0,
+            pending_batch_index: 0,
+            _padding: [0u8; 5],
+            expected_seq: 0,
+        };
```
ananas-block, line 212: zeroing out entries is expensive, what about adding a field that signals is_inserted or not?

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -440,34 +686,289 @@ impl<'a> BatchedMerkleTreeAccount<'a> {
...
+            entries_to_remove.push(entry);
+            self.remove_changelog_entries(&entries_to_remove);
+            
+            // For simplicity and safety, we won't recursively apply the changelog entry
+            // This is complex because we'd need to re-fetch the queue account
+            // but we can let the next update call handle it
+        }
+        
+        // 6. Return the batch append event.
+        Ok(MerkleTreeEvent::BatchAppend(event))
```
ananas-block, line 718: we need to return all events -> need to change return type to Vec<MerkleTreeEvent> for all update from queue methods.

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
+    }
+    
+    /// Remove changelog entries from the changelog
+    pub fn remove_changelog_entries(&mut self, entries: &[crate::changelog::BatchChangelog]) {
+        // Mark the entries as processed by zeroing them out
+        let zeroed_entry = crate::changelog::BatchChangelog {
+            old_root: [0u8; 32],
+            new_root: [0u8; 32],
+            leaves_hash_chain: [0u8; 32],
+            hash_chain_index: 0,
```
ananas-block, line 208: the fields `leaves_hash_chain` and `hash_chain_index` are likely not useful since we mix changelogs from input and output queue.

program-libs/batched-merkle-tree/src/queue_batch_metadata.rs
```diff
@@ -50,6 +50,10 @@ pub struct QueueBatches {
 impl QueueBatches {
     /// Returns the number of ZKP batches contained within a single regular batch.
     pub fn get_num_zkp_batches(&self) -> u64 {
+        // Prevent division by zero in test cases
+        if self.zkp_batch_size == 0 {
+            return 1; // Provide a minimal safe value for tests
```
ananas-block, line 55: this should panic in this case, but the case shouldn't be possible for checks during initialization.

program-libs/batched-merkle-tree/tests/account_access.rs
```diff
@@ -67,12 +67,15 @@ fn address_from_account_info() {
     account.lamports = merkle_tree_rent;
     // Test 1 functional init_batched_address_merkle_tree_from_account_info
     {
+        // Before initializing, set the data to a known state
+        account.data.fill(0);
+
```
ananas-block, line 72: not necessary

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -204,13 +375,25 @@ impl<'a> BatchedMerkleTreeAccount<'a> {
...
+        
+        // Parse or create changelog from remaining data
+        let changelog = if metadata.changelog_capacity > 0 && !remaining_data.is_empty() {
+            // Try to parse the changelog from the account data
+            ZeroCopyCyclicVecU64::<crate::changelog::BatchChangelog>::from_bytes(remaining_data)?
+        } else {
+            // If we can't parse, return an error - we shouldn't create temporary buffers
+            // in from_bytes as they won't outlive the function
+            return Err(ZeroCopyError::Size.into());
+        };
```
ananas-block, line 388: we should launch batched Merkle trees with a placeholder so that we can always assume that at least the length bytes exist in the account.
With this assumption we can just use `from_bytes_at` without any conditions to deserialize the changelog zero copy vec.

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
+        self.changelog.push(entry);
+    }
+    
+    /// Applies a changelog entry to update the tree state
+    pub fn apply_changelog_entry(&mut self, entry: &crate::changelog::BatchChangelog) 
+        -> Result<MerkleTreeEvent, BatchedMerkleTreeError> {
+        // Check if entry is applicable
+        if !self.is_changelog_entry_applicable(entry) {
+            return Err(BatchedMerkleTreeError::OldRootMismatch);
+        }
```
ananas-block, line 163: shouldn't throw an error just skip changelogs that are not applicable.

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
+    }
+    
+    /// Processes all applicable changelog entries
+    pub fn process_changelog_entries(&mut self) 
+        -> Result<Vec<MerkleTreeEvent>, BatchedMerkleTreeError> {
+        let mut events = Vec::new();
+        let mut applied_any = true;
+        
+        // Continue processing as long as we're making progress
```
ananas-block, line 231: it's ok to not apply any changelogs as well.

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -82,13 +97,169 @@ pub struct BatchedMerkleTreeAccount<'a> {
...
+        while applied_any {
```
ananas-block, line 232: the most efficient way is probably to filter changelogs for not inserted, expected seq to greater or equal, sort them ascendingly, and do all this in iterators so that we don't copy any state.
As soon as we sorted sequence numbers are not consecutive we can stop. 
We still need to check that the old root matches in every update.

program-libs/batched-merkle-tree/src/merkle_tree_metadata.rs
```diff
@@ -46,10 +46,24 @@ pub struct BatchedMerkleTreeMetadata {
...
+            num_batches: NUM_BATCHES as u64,
+            batch_size: TEST_DEFAULT_BATCH_SIZE,
+            bloom_filter_capacity: 20_000 * 8,
+            zkp_batch_size: TEST_DEFAULT_ZKP_BATCH_SIZE,
+            ..Default::default()
+        };
+        
+        // Default changelog capacity is 2x the number of zkp batches
+        let default_changelog_capacity = queue_batches.get_num_zkp_batches() * 2;
+
```
ananas-block, line 66: pls avoid unnecessary diff

ananas-block (commented):
Nice, I think it would be good to refactor the changelog specific functions into a separate file and add contained unit tests there to keep the diff in existing files minimal and make the changes easier to grasp.
Additionally, we can probably share the changelog logic between all the `update_tree_from_queue` methods.

PR #1698, opened 2025-04-17
perf: extend registered program pda account

ananas-block:
Issue:
- the account compression program performs an unnecessary pda derivation as a signer check

Solution:
- extend registered program pda account with the derived signer pda
- compare the stored signer pda stored in registered program pda account with the signer

Test Coverage:
1. functional
2. failing:
    2.1. resize again
    2.2. account invalid owner
    2.3. account invalid discriminator


sdk-libs/program-test/src/env_accounts_v1.rs
```diff
@@ -0,0 +1,37 @@
+// This file is generated by getAccountState.sh. Do not edit it manually.
+
+use std::str::FromStr;
+
+use solana_sdk::{account::Account, pubkey::Pubkey};
+pub fn get_registered_program_pda() -> Account {
```
ananas-block, line 6: keep v1 accounts for testing.

sergeytimoshin (approved):

PR #1710, opened 2025-04-24
refactor: counter program 2

ananas-block:
### Changes
1. restore anchor counter example
2. restore sdk-anchor-test
3. remove solana program dependency from program libs crates
4. use updated cargo generate template in cli
5.  sdk:
    5.1. add `CompressionInstruction` as input type for  `verify_compressed_account_infos`
    5.2. remove `LightInstructionData`
    5.3. add `ValidityProof` wraps `Option<CompressedProof>` to encourage the pattern of using `Option<Proof>`
6. `CompressionCpiAccounts` -> `CpiAccounts`
7. `CompressionCpiInputs` -> `CpiInputs`
8. `verify_compressed_account_infos` -> `invoke_light_system_program`
9. new input parameters for `invoke_light_system_program`

```
pub fn invoke_light_system_program(
    invoking_program_id: &Pubkey,
    account_infos: &[AccountInfo],
    instruction: Instruction,
)
```

10. add  method `invoke_light_system_program` to `CpiInputs`
    - the purpose is to hide `invoking_program_id` which is confusing for devs used to `solana_program::invoke` and `solana_program::instruction`
    since instruction requires the program id that is being invoked but `invoke_light_system_program` requires the program id of the invoking program to derive the signer
11. add `create_instruction` function to create a `light_system_program` `invoke_cpi` instruction


prover/client/src/gnark/helpers.rs
```diff
@@ -81,54 +81,57 @@ pub struct ProverConfig {
 }
 
 pub async fn spawn_prover(restart: bool, config: ProverConfig) {
-    if let Some(_project_root) = get_project_root() {
-        let prover_path: &str = {
-            #[cfg(feature = "devenv")]
```
ananas-block, line 86: revert

sdk-libs/program-test/src/indexer/test_indexer.rs
```diff
@@ -1192,6 +1212,7 @@ where
         }
     }
 
+    #[cfg(feature = "devenv")]
```
ananas-block, line 1215: rename feature to v2

sdk-libs/sdk/src/lib.rs
```diff
@@ -15,17 +15,56 @@ pub mod utils;
...
-};
+use light_verifier::CompressedProof;
+use solana_account_info::AccountInfo;
+use solana_cpi::invoke_signed;
+use solana_instruction::{AccountMeta, Instruction};
+use solana_msg::msg;
+use solana_program_error::ProgramError;
+use solana_pubkey::{pubkey, Pubkey};
+
+pub type AddressProof = CompressedProof;
```
ananas-block, line 31: Tradeoff `AddressProof` vs `ValidityProof`

if we return `ValidityProof` from rpc `AddressProof` doesn't make sense.

ananas-block:
@ghulamMustafaRaza everything that is not behind a v2 feature should be compatible with deployed light programs.
I will double check that before the release.

SwenSchaeferjohann (approved):

PR #1715, opened 2025-04-29
fix: check_account_balance_is_rent_exempt return rent_exemption

ananas-block:
Issue:
- an increase in accounts lamports increased the rollover fee

Solution:
- return rent exempt amount instead of account lamports


SwenSchaeferjohann (approved):

program-tests/account-compression-test/tests/batched_merkle_tree_test.rs
```diff
@@ -4,7 +4,10 @@ use account_compression::{errors::AccountCompressionErrorCode, ID};
 use anchor_lang::{
     error::ErrorCode, prelude::AccountMeta, AnchorSerialize, InstructionData, ToAccountMetas,
 };
-use anchor_spl::token::Mint;
+use anchor_spl::{
+    token::Mint,
+    token_2022::spl_token_2022::extension::default_account_state::instruction::decode_instruction,
+};
```
ananas-block, line 10:
```suggestion
use anchor_spl::token::Mint;
```

PR #1718, opened 2025-04-30
feat: add batch compress tokens

ananas-block:
### Changes:
1. add const generic `IS_MINT_TO` to `process_mint_to` 
2. use  `IS_MINT_TO` to either `mint_spl_to_pool_pda` or transfer spl tokens to a token pool
3. introduce `PubkeyTrait` to enable zero copy instruction data for the batch compress instruction
4. introduce `ZeroCopyNumTrait` to enable zero copy instruction data for the batch compress instruction and backward compatibility of `mint_spl_to_pool_pda`, and `create_output_compressed_accounts`.
5. refactor `spl_token_transfer` to manually match the token program id for CU savings
6. rename `add_token_data_to_input_compressed_accounts` -> `add_data_hash_to_input_compressed_accounts`

### Test Coverage:
1. Functional compress 1 to 26 recipients
2. Failing unequal recipients amounts len
3. Failing insufficient balance
4. Failing sender account and token pool account with different mint
5. Failing invalid derived token pool pda
10. Failing invalid token pool pda derived from different index
11. Failing no sender token account

## Performance Optimizations

1. remove unnecessary anchor seed check
    The cpi authority is implicitly validated when it signs the cpi.
2. replace system program cpi with manual cpi to new `invoke_cpi_with_readonly` instruction
    2.1. refactor `add_data_hash_to_input_compressed_accounts` to use `InAccount` instead of  `PackedCompressedAccountWithMerkleContext`
3. add `with_transaction_hash` to `transaction` instruction data
    3.1. extend instruction data by 5 bytes for backwards compatibility since borsh ignores trailing bytes deserialization will not fail in case complete instruction data is provided


SwenSchaeferjohann (approved):

programs/compressed-token/src/instructions/create_token_pool.rs
```diff
@@ -110,7 +113,24 @@ pub struct AddTokenPoolInstruction<'info> {
...
+}
+
+#[inline(always)]
+pub fn check_spl_token_pool_derivation_with_index(
+    token_pool_pda: &Pubkey,
+    mint: &Pubkey,
+    index: u8,
+    bump: Option<u8>,
+) -> Result<()> {
+    msg!("bump {:?}", bump);
```
ananas-block:
```suggestion
```

ananas-block:
```suggestion
```

PR #1725, opened 2025-05-04
chore: review 1718

SwenSchaeferjohann:


programs/compressed-token/src/process_mint.rs
```diff
@@ -84,22 +84,18 @@ pub fn process_mint_to<'info, const IS_MINT_TO: bool>(
             for a in amounts {
                 amount += (*a).into();
             }
-            let index = if let Some(index) = index {
-                index
-            } else {
-                panic!("No index provided for batch compress.");
-            };
+            let index = index.ok_or_else(|| panic!("No index provided for batch compress."))?;
```
SwenSchaeferjohann, line 87: may have to replace with `ok_or`

ananas-block (approved):

PR #1747, opened 2025-05-13
ci: change runner to ubuntu-latest for forester tests

sergeytimoshin:


ananas-block:
did't work because ubuntu-latest  doesn't have enough ram

PR #1752, opened 2025-05-15
chore: use remote prover server on CI tests

sergeytimoshin:
* replace SERVER_ADDRESS constant with get_server_address()
* if PROVER_URL is set do not spawn local prover


ananas-block (approved):
Nice solution.
How do we restrict access to the prover server?

wrt to the failing test `context: "get_validity_proof",` might be the required proving key doesn't exit.  `ForesterTest` mode didn't include combined provingkeys until one of my last prs.

sergeytimoshin:
> How do we restrict access to the prover server?

For now, I've just hidden it behind [github env secrets](https://github.com/Lightprotocol/light-protocol/settings/secrets/actions), but even a rudimentary auth wouldnt be out of place, I'll think about it.

ananas-block:
could we restrict it to an ip?

PR #1753, opened 2025-05-16
feat: regenerate photon-api crate & v2 proof handling

sergeytimoshin:
* photon dependency in scripts/install.sh has been updated to a newer commit
* photon-api crate has been regenerated
* get_validity_proof_v2 updated accordingly
* CompressedProofWithContextV2 has been refactored, introducing new associated structs like AccountProofInputs, AddressProofInputs, RootIndex, MerkleContextV2, and TreeContextInfo to better represent the V2 proof context.

Forester Tests:
* address_v2_test.rs has been updated, the logic for obtaining validity proofs was updated to directly use the Photon indexer


sdk-libs/photon-api/src/models/account_context.rs
```diff
@@ -44,7 +44,7 @@ impl AccountContext {
         nullified_in_tree: bool,
         queue: String,
         spent: bool,
-        tree_type: u16,
+        tree_type: i32,
```
ananas-block, line 47: is this on purpose?

sdk-libs/photon-api/src/models/compressed_proof_with_context_v2.rs
```diff
@@ -14,33 +14,73 @@ use crate::models;
...
+    pub leaf_index: u64,
+    pub merkle_context: MerkleContextV2,
+}
+
+#[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AddressProofInputs {
+    pub address: String,
+    pub root: String,
+    pub root_index: RootIndex,
```
ananas-block, line 40: this should be u16 right?
(Since address proofs don't work with proof by index.)

sdk-libs/photon-api/src/models/account_context.rs
sergeytimoshin: Fixed it, thanks.

ananas-block (approved):

PR #1761, opened 2025-05-23
refactor: redis-based job queue for prover

sergeytimoshin:


forester-utils/src/instructions/state_batch_nullify.rs
```diff
@@ -49,11 +44,11 @@ pub async fn create_nullify_batch_ix_data<R: RpcConnection, I: Indexer>(
...
         let batch = &merkle_tree.queue_batches.batches[batch_idx];
         let num_inserted_zkps = batch.get_num_inserted_zkps();
-        let num_current_zkp = batch.get_current_zkp_batch_index();
-        let num_ready_zkps = num_current_zkp.saturating_sub(num_inserted_zkps);
+        let num_current_zqp = batch.get_current_zkp_batch_index();
+        let num_ready_zkps = num_current_zqp.saturating_sub(num_inserted_zkps);
 
         let mut leaves_hash_chains = Vec::new();
-        for i in num_inserted_zkps..num_current_zkp {
+        for i in num_inserted_zkps..num_current_zqp {
```
ananas-block, line 51: zqp -> zkp ?

forester-utils/src/proof_client.rs
```diff
@@ -0,0 +1,285 @@
...
+use light_prover_client::gnark::{
+    constants::PROVE_PATH,
+    proof_helpers::{compress_proof, deserialize_gnark_proof_json, proof_from_json_struct},
+};
+use reqwest::Client;
+use serde::Deserialize;
+use tokio::time::sleep;
+use tracing::{debug, error, info, warn};
+
+use crate::error::ForesterUtilsError;
```
ananas-block, line 13: could it make sense to move this into `light-prover-client` ?
Or do we want to keep `light-prover-client`  as a request wrapper for the go prover server?

program-tests/compressed-token-test/tests/test.rs
```diff
@@ -1247,7 +1246,6 @@ async fn test_transfers() {
     spawn_prover(ProverConfig {
         run_mode: Some(ProverMode::ForesterTest),
         circuits: vec![],
-        restart: false,
```
ananas-block, line 1250: why remove the restart field?
(I use this during debugging to reduce repeated test time and we use it in program tests to make CI pass else CI testing time is too long.)

scripts/install.sh
```diff
@@ -185,6 +186,151 @@ install_dependencies() {
     fi
 }
 
+
+install_redis() {
```
SwenSchaeferjohann, line 190: formatting seems off in this file

.github/workflows/rust.yml
```diff
@@ -16,7 +16,6 @@ on:
       - "**/Cargo.*"
       - "prover/client/**"
       - ".github/workflows/rust.yml"
-    types: [opened, synchronize, reopened, ready_for_review]
```
SwenSchaeferjohann, line 19: why remove this?

SwenSchaeferjohann (commented):
LGTM on first pass!
 
added some questions below. also, is there a better way to handle the Redis port throughout the repo?

.github/workflows/rust.yml
sergeytimoshin: It's basically the same behaviour, just simplification, but lets make explicit

SwenSchaeferjohann (approved):

PR #1763, opened 2025-05-29
chore: add c-token wrappers for example-compressed-claim

SwenSchaeferjohann:
#1603, but rebased to the latest main

### DIFF
--- a/sdk-libs/compressed-token-client/src/instructions.rs
+++ b/sdk-libs/compressed-token-client/src/instructions.rs
@@ -164,7 +164,7 @@ pub fn create_decompress_instruction(
             lamports: None,
             merkle_tree: params
                 .output_state_tree
-                .unwrap_or(merkle_contexts[0].merkle_tree_pubkey),
+                .unwrap_or(merkle_contexts[0].merkle_tree_pubkey.into()),
         }]
     } else {
         vec![]


sdk-libs/compressed-token-sdk/src/cpi/accounts.rs
```diff
@@ -0,0 +1,20 @@
+use solana_program::account_info::AccountInfo;
+
+/// CPI Accounts for decompressing compressed token accounts.
+pub struct CompressedTokenDecompressCpiAccounts<'a> {
```
ananas-block, line 4: The way to pass accounts here is quite different from how we do it `light-sdk` [see docs.rs](https://github.com/Lightprotocol/light-protocol/blob/0eee5b8aa3f49b662f8d3bc935b5e59ec73d8c28/sdk-libs/sdk/src/cpi/accounts.rs#L7).

We also need helpers to create accounts in the client.
For the client it probably makes sense to use the same pattern of `PackedAccounts` here.
It might make sense to extend `PackedAccounts` [see github](https://github.com/Lightprotocol/light-protocol/blob/0eee5b8aa3f49b662f8d3bc935b5e59ec73d8c28/sdk-libs/sdk/src/instruction/pack_accounts.rs#L9).
This is a [ts implementation of PackedAccounts](https://github.com/ananas-block/test-compressed-program/blob/ffe3d20a45c0a86f8fdb1dd6577838b64a774823/tests/test.ts#L309).

PR #1765, opened 2025-05-30
fix: indexer get_compressed_account return type

ananas-block:
### Changes:
1. add `get_compressed_accounts_by_owner`
2. remove vector in base58 to fixed sized array conversion
3. implement `Account` type equivalent to photon indexer `AccountV2`
4. use new `Account` type as return type for `get_compressed_account`


sdk-libs/client/src/indexer/types.rs
```diff
@@ -114,7 +119,7 @@ pub struct ProofRpcResultV2 {
     pub address_root_indices: Vec<u16>,
 }
 
-#[cfg(feature = "v2")]
+// #[cfg(feature = "v2")]
```
sergeytimoshin, line 122: leftover?

PR #1768, opened 2025-05-31
refactor: light client

ananas-block:
### Issue:
- we don't have a complete rust client implementation

### Changes:
1. Expose retry config with every Indexer trait method
    1.1.  allow retries based on a slot to wait until the indexer has caught up to the desired slot
    using a timestamp would be more elegant then we don't need to do another rpc call to get the current slot.
2. Indexer trait resembles almost the complete photon api with consistent input and return types
    2.1. omitted methods are:
         1. `getLatestCompressionSignatures`
         2. `getLatestNonVotingSignatures`
    2.2. return types are V2 return types
    2.3. methods with a V2 implemenation use the v2 api method if compiled with v2 feature


sdk-libs/client/src/indexer/indexer_trait.rs
```diff
@@ -0,0 +1,188 @@
...
+    ) -> Result<Response<u64>, IndexerError>;
+
+    async fn get_multiple_compressed_accounts(
+        &self,
+        addresses: Option<Vec<Address>>,
+        hashes: Option<Vec<Hash>>,
+        config: Option<IndexerRpcConfig>,
+    ) -> Result<Response<Vec<Account>>, IndexerError>;
+
+    async fn get_compressed_token_balances_by_owner(
```
SwenSchaeferjohann: helius actually has a getCompressedTokenBalancesByOwnerV2 endpoint that has a different response than the, now deprecated, getCompressedTokenBalancesByOwner endpoint

sdk-libs/client/src/indexer/indexer_trait.rs
```diff
@@ -0,0 +1,188 @@
...
+    ) -> Result<ResponseWithCursor<Vec<Account>, [u8; 32]>, IndexerError>;
+
+    async fn get_compressed_account(
+        &self,
+        address: Option<Address>,
+        hash: Option<Hash>,
+        config: Option<IndexerRpcConfig>,
+    ) -> Result<Response<Account>, IndexerError>;
+
+    async fn get_compressed_token_accounts_by_owner(
```
SwenSchaeferjohann: we'll need cursor for
- get_compressed_token_accounts_by_owner
- get_compressed_token_accounts_by_delegate
- get_compressed_token_balances_by_owner (deprecated, so not strictly needed imo)
- get_compressed_token_balances_by_owner_v2

for example, in JS, we use the same optional param for all of them:
```typescript 
async getCompressedTokenAccountsByOwner(
        owner: PublicKey,
        options?: GetCompressedTokenAccountsByOwnerOrDelegateOptions,
)
```
```typescript
export interface GetCompressedTokenAccountsByOwnerOrDelegateOptions {
    mint?: PublicKey;
    cursor?: string;
    limit?: BN;
}
```
link [here](https://github.com/Lightprotocol/light-protocol/blob/main/js/stateless.js/src/rpc-interface.ts#L128).

we also need cursor for:
- get_compressed_accounts_by_owner
- getCompressionSignaturesForAddress
- getCompressionSignaturesForOwner
- getCompressionSignaturesForTokenOwner
- getCompressedMintTokenHolders

example consumer of this will be the Swap Aggregators' rust-based indexers

sdk-libs/client/src/indexer/indexer_trait.rs
SwenSchaeferjohann: Re `get_compressed_token_account_balances_v2` - I think we should keep the explicit _v2 endpoint so as to not break existing apps. if you want to keep only one, it's fine to just keep _v2 imo.

sdk-libs/client/src/indexer/photon_indexer.rs
```diff
@@ -223,94 +237,49 @@ impl Indexer for PhotonIndexer {
...
+                        cursor: options.as_ref().and_then(|o| o.cursor.clone()),
+                        data_slice: options.as_ref().and_then(|o| {
+                            o.data_slice.as_ref().map(|ds| {
+                                Box::new(photon_api::models::DataSlice {
+                                    length: ds.length as u32,
+                                    offset: ds.offset as u32,
+                                })
+                            })
+                        }),
+                        filters: None, // TODO: Convert filters if needed
```
ananas-block: use filter

sdk-libs/client/src/indexer/photon_indexer.rs
```diff
@@ -321,154 +290,80 @@ impl Indexer for PhotonIndexer {
...
```
ananas-block: use filter

SwenSchaeferjohann (approved):

PR #1775, opened 2025-06-04
feat: light client get cached merkle tree

ananas-block:
### Changes:
1. rename trait `RpcConnection` -> `Rpc`
2. rename `RpcConnectionConfig` ->  `RpcConfig`
3. rename `SolanaRpcConnection`  -> `LightClient`
4. rename `SolanaRpcUrl` -> `RpcUrl`
5. add `pack_tree_accounts` method to `ValidityProofWithContext` see counter example for usage
6. add `get_output_tree_index` method if V1 tree pack Merkle tree, if V2 tree pack output queue.
7. feat cached Merkle trees
    7.1.  `Rpc::new` optional fetch Merkle trees with client instantiation
    7.2. add get trees functions to `Rpc` trait (impl these for LightClient, LightProgramTest)
    ```
    /// Fetch the latest state tree addresses from the cluster.
    async fn get_latest_active_state_trees(&mut self) -> Result<Vec<MerkleContext>, RpcError>;

    /// Gets active state trees.
    /// State trees are cached and have to be fetched or set.
    fn get_active_state_trees(&self) -> Vec<MerkleContext>;

    /// Gets a random active state tree.
    /// State trees are cached and have to be fetched or set.
    fn get_random_active_state_tree(&self) -> MerkleContext;

    fn get_address_tree_v1(&self) -> MerkleContext
    ```
    7.3. add state tree and nullifier lookup table utils:
        `get_light_state_tree_infos`
         `default_state_tree_lookup_tables`


sdk-libs/client/src/indexer/types.rs
```diff
@@ -150,7 +152,72 @@ impl AddressProofInputs {
     }
 }
 
+pub struct PackedMerkleTreeAccounts {
+    pub input_merkle_contexts: Vec<PackedMerkleContext>,
+    pub output_merkle_tree_indices: Vec<u8>,
```
ananas-block: I am considering two alternatives:

1. return only one `output_merkle_tree_index`.
```
pub struct PackedMerkleTreeAccounts {
    pub input_merkle_contexts: Vec<PackedMerkleContext>,
    pub output_merkle_tree_index:u8,
    pub new_address_merkle_contexts: Vec<PackedAddressMerkleContext>,
}
```

2. use the account info schema
```
pub struct PackedMerkleTree {
    in: PackedMerkleContext,
    out: u8,
}

pub struct PackedMerkleTreeAccounts {
    pub merkle_contexts: Vec<PackedMerkleTree>,
    pub new_address_merkle_contexts: Vec<PackedAddressMerkleContext>,
}
```

Pro 1:
- it clearly exposes the types that are used in `CompressedAccountMeta`

Pro 2:
- doesn't push users to use only one tree for all outputs (although fewer trees are good)

Leaning towards 1.

examples/anchor/counter/tests/test.rs
SwenSchaeferjohann, line 1: In `rpc.test_accounts.v1_address_trees` we can store `Vec<AddressMerkleContext>` instead of `Vec<AddressMerkleTreeAccounts>` (it's just different field names)

this would simplify

```rust
let address_merkle_context = AddressMerkleContext {
        address_merkle_tree_pubkey: rpc.test_accounts.v1_address_trees[0].merkle_tree,
        address_queue_pubkey: rpc.test_accounts.v1_address_trees[0].queue,
    };
```

into:

```rust
let address_merkle_context = rpc.test_accounts.v1_address_trees[0];

```

examples/anchor/counter/tests/test.rs
```diff
@@ -182,16 +174,16 @@ where
...
         .value;
 
-    let output_merkle_tree_index = remaining_accounts.insert_or_get(output_merkle_tree);
-    let packed_address_merkle_context = pack_address_merkle_context(
-        &address_merkle_context,
-        &mut remaining_accounts,
-        rpc_result.get_address_root_indices()[0],
-    );
+    let output_merkle_tree_index = rpc
+        .get_random_active_state_tree()
```
SwenSchaeferjohann: I think a big gain would come from renaming the new `MerkleContext` into `StateTreeInfo` and have the OG `MerkleContext` simply include `StateTreeInfo`. This would then be consistent with both the photon response and the TS sdk.
here's how we do it in TS ([code](https://github.com/Lightprotocol/light-protocol/blob/b9d6ee9ab28f95f82add8c0dd7b744c7015a985b/js/stateless.js/src/state/types.ts#L68-L90)):
```typescript
export type MerkleContext = {
    /**
     * Poseidon hash of the account. Stored as leaf in state tree
     */
    hash: BN;
    /**
     * Position of `hash` in the State tree
     */
    leaf_index: u32;
    /**
     * Whether the account can be proven by index or by merkle proof
     */
    prove_by_index: boolean;
   /**
     * Tree info
     */
    tree_info: StateTreeInfo;
};
```

```typescript
export type StateTreeInfo = {
    /**
     * Pubkey of the tree account.
     */
    tree: PublicKey;
    /**
     * Pubkey of the queue account associated with the tree.
     */
    queue: PublicKey;
    /**
     * The type of tree. One of {@link TreeType}.
     */
    tree_type: TreeType;
    /**
     * Optional compressed cpi context account.
     */
    cpi_context: Option<PublicKey>;
    /**
     * Next tree info. Is `some` if the next tree should be used for the next
     * state transition.
     */
    next_tree_info: Option<StateTreeInfo>;
};
```

in rust we currently have a separate type `TreeContextInfo` which almost resembles the rust merkle_context. This could then simply be removed in favor of using `StateTreeInfo` here too.
```rust
#[derive(Clone, Copy, Default, Debug, PartialEq)]
pub struct TreeContextInfo {
    pub cpi_context: Option<Pubkey>,
    pub queue: Pubkey,
    pub tree: Pubkey,
    pub tree_type: TreeType,
}
```

SwenSchaeferjohann: along with this, `getRandomActiveStateTree` is deprecated in TS so may confuse users - I'd suggest to rename to `getRandomStateTreeInfo`

examples/anchor/counter/tests/test.rs
ananas-block: added `get_address_tree_v1` to `Rpc` trait it returns a `MerkleContext` like this:
``` 
let address_merkle_context = rpc.get_address_tree_v1();
```

ananas-block: wrt changing the underlying datatype, I am bit wary of this because it might incur a lot of changes in the program test code.
Ideally devs shouldn't access the internal types and use `get_address_tree_v1` instead which is available both in `LightClient` and `LightProgramTest`.

examples/anchor/counter/tests/test.rs
```diff
@@ -182,16 +174,16 @@ where
...
+        .get_output_tree_index(&mut remaining_accounts)?;
+    let address_merkle_context = rpc_result
+        .pack_merkle_tree_accounts(&mut remaining_accounts)
+        .new_address_merkle_contexts[0];
```
SwenSchaeferjohann: have you considered:

```rust
let address_merkle_context = pack_merkle_contexts(&mut remaining_accounts, &rpc_result.accounts, &rpc_result.addresses).new_address_merkle_contexts[0];
```

it's more clear what's happening underneath, it's flexible, and keeps the concerns of validityProof and account packing separate

examples/anchor/counter/tests/test.rs
ananas-block: The `MerkleContext` struct you are referencing has the same layout as `StateTreeInfo`.
Hm so you are essentially proposing to rename `MerkleContext` to `StateTreeInfo` and `AddressTreeInfo`.
This `MerkleContext` has a tree type field so it would be  be more coherent to name it `TreeInfo`.

> in rust we currently have a separate type TreeContextInfo which almost resembles the rust merkle_context. This could then simply be removed in favor of using StateTreeInfo here too.

The reason `TreeContextInfo` exists is that it doesn't have `next_tree_info` which should always be `None`.
Happy to rename it to `NextTreeInfo` lol.
Also happy to rename `MerkleContext` to `TreeInfo`.

examples/anchor/counter/tests/test.rs
ananas-block: Which part would make what's underneath more clear, the separate function or separate inputs?


> `pack_merkle_contexts(&mut remaining_accounts, &rpc_result.accounts, &rpc_result.addresses)`

Two separate rpc_result inputs don't make sense imo because this function will only be used by outputs from the rpc result. The accounts and addresses types are only used in `ValidityProofWithContext`.

Using it as a method makes it easier to find imo.

ananas-block:
rebased on https://github.com/Lightprotocol/light-protocol/pull/1776

examples/anchor/counter/tests/test.rs
SwenSchaeferjohann: I see, the reasoning being `root_index` is part of `PackedAddressContext` but only available after `get_validity_proof`.

have you considered adding `root_index` to `PackedMerkleContext`? then the DX would be consistent across address and account.

(in this case, we should still keep the unpacked `MerkleContext` without `root_index`, to keep it non-expiring.)

ananas-block: > have you considered adding root_index to PackedMerkleContext? then the DX would be consistent across address and account.

Yes I am currently adding:
```
pub struct PackedTreeInfo {
    pub root_index: u16,
    pub prove_by_index: bool,
    pub merkle_tree_pubkey_index: u8,
    pub queue_pubkey_index: u8,
    pub leaf_index: u32,
}
```

I am also looking to rename `merkle_context` -> `tree_info` in `light-sdk`

sdk-libs/client/src/indexer/types.rs
```diff
@@ -145,12 +150,88 @@ impl AddressProofInputs {
...
+pub struct PackedAccountTreeInfos {
+    pub packed_tree_infos: Vec<PackedTreeInfo>,
+    pub output_tree_index: u8,
+}
+
+#[derive(Clone, Default, Debug, PartialEq)]
+pub struct PackedTreeAccounts {
+    pub account_trees: Option<PackedAccountTreeInfos>,
+    pub address_trees: Vec<PackedAddressMerkleContext>,
+}
```
ananas-block: `PackedTreeAccounts` -> `PackedTreeInfos`
`PackedAddressMerkleContext` -> `PackedAddressTreeInfo`
`PackedAccountTreeInfos` -> `PackedStateTreeInfos`

examples/anchor/memo/tests/test.rs
```diff
@@ -183,7 +183,7 @@ where
         .await
         .unwrap();
 
-    let address_merkle_context = AddressMerkleContext {
+    let address_merkle_context = AddressTreeInfo {
         address_merkle_tree_pubkey: env.v1_address_trees[0].merkle_tree,
```
SwenSchaeferjohann, line 187: how disruptive would it be to also switch `address_merkle_tree_pubkey` -> `address_tree_pubkey`
(and `merkle_tree_pubkey` -> `state_tree_pubkey`, respectively)

examples/anchor/counter/tests/test.rs
```diff
@@ -173,26 +128,25 @@ where
...
 
-    let output_merkle_tree_index = remaining_accounts.insert_or_get(output_merkle_tree);
-    let packed_address_merkle_context = pack_address_merkle_context(
-        &address_merkle_context,
-        &mut remaining_accounts,
-        rpc_result.get_address_root_indices()[0],
-    );
+    let output_tree_index = rpc
+        .get_random_state_tree_info()
+        .get_output_tree_index(&mut remaining_accounts)?;
```
SwenSchaeferjohann: it may be helpful to name `get_output_tree_index` -> `get_output_state_tree_index` and `output_tree_index`  -> `output_state_tree_index`

especially for devs dealing with address_trees and state_trees in the same instruction

thoughts?

examples/anchor/counter/tests/test.rs
```diff
@@ -219,42 +173,35 @@ where
...
-        .unwrap()
+        .get_validity_proof(vec![hash], vec![], None)
+        .await?
         .value;
 
-    let packed_merkle_context =
-        pack_merkle_context(&compressed_account.merkle_context, &mut remaining_accounts);
+    let packed_tree_accounts = rpc_result
+        .pack_tree_infos(&mut remaining_accounts)
+        .account_trees
```
SwenSchaeferjohann: for consistency I recommend we use `state_trees` instead of `account_trees`.

(or `state_tree_infos` and `address_tree_infos`, or `packed_state_tree_infos` and `packed_address_tree_infos`)

sdk-libs/client/src/indexer/types.rs
```diff
@@ -276,50 +357,81 @@ impl ValidityProofWithContext {
...
+    pub fn get_output_tree_index(
+        &self,
+        packed_accounts: &mut PackedAccounts,
+    ) -> Result<u8, IndexerError> {
+        match self.tree_type {
+            TreeType::StateV1 => Ok(packed_accounts.insert_or_get(self.tree)),
+            TreeType::StateV2 => Ok(packed_accounts.insert_or_get(self.queue)),
+            _ => Err(IndexerError::InvalidPackTreeType),
+        }
+    }
```
SwenSchaeferjohann: by it's name, i'd expect this to really just get the index (akin to what `get_output_tree_index` does in CompresssedAccountMetaTrait)

thoughts about naming this method here `get_output_tree_index` -> `pack_output_tree_index` or even `pack_output_state_tree`?

examples/anchor/counter/tests/test.rs
ananas-block: changed it, it is more clear.

We have some other `light-sdk` types where we can think about adapting field names as well such as:

```
pub struct CompressedAccountMetaNoLamportsNoAddress {
    pub tree_info: PackedStateTreeInfo,
    pub output_tree_index: u8,
}
```

Could turn into:


```
pub struct CompressedAccountMetaNoLamportsNoAddress {
    pub state_tree_info: PackedStateTreeInfo,
    pub output_state_tree_index: u8,
}
```
Pro:
- more clear
Con:
- longer field names are annoying I am trying to make them shorter where ever I can
- for tree_info the type already includes the word state

Mixed option:

```
pub struct CompressedAccountMetaNoLamportsNoAddress {
    pub tree_info: PackedStateTreeInfo,
    pub output_state_tree_index: u8,
}
```

ananas-block: Other occurences:
1. `CompressedAccountMetaTrait` has a getter `get_output_tree_index`
2. `AccountInfoTrait` -> input values `output_tree_index`
3. multiple occurences in `light-compressed-account` of `output_merkle_tree_index` in `InAccountInfo` and `OutAccountInfo`

SwenSchaeferjohann (approved):

PR #1789, opened 2025-06-09
refactor: resync photon-api with photon

sergeytimoshin:


sdk-libs/client/src/indexer/photon_indexer.rs
```diff
@@ -365,217 +339,125 @@ impl Indexer for PhotonIndexer {
         .await
     }
 
-    async fn get_compressed_account(
+    async fn get_compressed_balance(
         &self,
-        address: Address,
+        address: Option<Address>,
+        hash: Option<Hash>,
         config: Option<IndexerRpcConfig>,
```
ananas-block, line 346: `get_compressed_account` is split up into `get_compressed_account(address: , ..)` and `get_compressed_account_by_hash(hash: ,..)`

It seems like the order of implementations is different here, in zed you can order the method implementations in the same way as the trait definition. Can you order it maybe that reduces the diff?

sergeytimoshin: > you can order the method implementations in the same way as the trait definition

This is exactly what it did, it was unsorted before.
Trait: https://github.com/Lightprotocol/light-protocol/blob/sergey/photon-api-bump/sdk-libs/client/src/indexer/indexer_trait.rs#L20-L31
Photon indexer: https://github.com/Lightprotocol/light-protocol/blob/sergey/photon-api-bump/sdk-libs/client/src/indexer/photon_indexer.rs#L159-L198

PR #1795, opened 2025-06-11
fix: conflicting pinocchio & solana crates compilation errors

ananas-block:


ananas-block:
This pr should be merged with rebase and merge intead of squash in merge to persist the commits!

ananas-block:
consider to move `AccountInfoTrait` into a separate crate

program-libs/account-checks/src/checks.rs
```diff
@@ -98,8 +90,9 @@ pub fn check_discriminator<T: Discriminator>(bytes: &[u8]) -> Result<(), Account
 }
 
 /// Checks that the account balance is greater or eqal to rent exemption.
-pub fn check_account_balance_is_rent_exempt(
-    account_info: &AccountInfo,
+#[cfg(any(feature = "pinocchio", feature = "solana"))]
```
SwenSchaeferjohann: any reason why we use conditional compilation explicitly here but not for the other functions?

program-libs/merkle-tree-metadata/src/rollover.rs
```diff
@@ -67,11 +67,18 @@ impl RolloverMetadata {
...
         #[cfg(target_os = "solana")]
         {
-            use crate::{Clock, Sysvar};
-            self.rolledover_slot = Clock::get().unwrap().slot;
+            #[cfg(feature = "pinocchio")]
+            {
+                use pinocchio::sysvars::{clock::Clock, Sysvar};
+                self.rolledover_slot = Clock::get().unwrap().slot;
+            }
+            #[cfg(not(feature = "pinocchio"))]
```
SwenSchaeferjohann, line 77: same as?
```suggestion
            #[cfg(feature = "solana")]
```

program-libs/account-checks/src/account_info/pinocchio.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn find_program_address(_seeds: &[&[u8]], _program_id: &[u8; 32]) -> ([u8; 32], u8) {
+        #[cfg(target_os = "solana")]
+        {
+            let program_pubkey = pinocchio::pubkey::Pubkey::from(*_program_id);
+            let (pubkey, bump) = pinocchio::pubkey::find_program_address(_seeds, &program_pubkey);
+            (pubkey, bump)
+        }
+        // Pinocchio does not support find_program_address outside of target_os solana.
+        // That is annoying for rust unit tests.
+        #[cfg(all(not(target_os = "solana"), feature = "solana"))]
```
SwenSchaeferjohann, line 54: is `feature = "solana"` possible using `pinocchio.rs`?

program-libs/account-checks/src/account_info/pinocchio.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    fn find_program_address(_seeds: &[&[u8]], _program_id: &[u8; 32]) -> ([u8; 32], u8) {
+        #[cfg(target_os = "solana")]
+        {
+            let program_pubkey = pinocchio::pubkey::Pubkey::from(*_program_id);
+            let (pubkey, bump) = pinocchio::pubkey::find_program_address(_seeds, &program_pubkey);
+            (pubkey, bump)
+        }
+        // Pinocchio does not support find_program_address outside of target_os solana.
+        // That is annoying for rust unit tests.
```
SwenSchaeferjohann:
```suggestion
```

program-libs/account-checks/src/checks.rs
ananas-block: good catch its a letftover

program-libs/merkle-tree-metadata/src/rollover.rs
ananas-block: in this case we need this condition else if we compile with all features we get an import conflict because we are importing this twice `use pinocchio::sysvars::{clock::Clock, Sysvar};' and we only need it once.

SwenSchaeferjohann (approved):
LGTM

PR #1796, opened 2025-06-11
fix: circuit selection for forester mode

sergeytimoshin:


prover/server/prover/proving_keys_utils.go
```diff
@@ -158,24 +153,21 @@ func GetKeys(keysDir string, runMode RunMode, circuits []string) []string {
...
 	switch runMode {
-	case Forester: // inclusion + non-inclusion
+	case Forester: // inclusion + non-inclusion + append + update + address-append
 		keys = append(keys, inclusionKeys...)
 		keys = append(keys, nonInclusionKeys...)
 		keys = append(keys, appendWithProofsKeys...)
 		keys = append(keys, updateKeys...)
-	case ForesterTest: // append-test + update-test + address-append-test
+		keys = append(keys, addressAppendKeys...)
+	case ForesterTest: // inclusion + non-inclusion + append-test + update-test + address-append-test
```
ananas-block:
```suggestion
	case ForesterTest: // inclusion + non-inclusion + combined + append-test + update-test + address-append-test
```

ananas-block (approved):
One nit, It is probably ok that forester mode doesn't include combined proving keys, we should just keep it in mind. ForesterTest didn't include combined proving keys at first, that took me a while to figure out.

PR #1797, opened 2025-06-11
chore: address-append-250 & combined_26_40 download fix

sergeytimoshin:
* regenerate address_append_40_250, add to DO storage
* update address_append_40_250 vkey
* add address_append_40_250 to download_keys.sh (full mode)
* add combined_26_40* to download_keys.sh
* sort CHECKSUM entries when generating checksum


ananas-block:
We don't support combined proofs of 26&40 tree heights onchain.
Probably best to remove that proving key.

PR #1809, opened 2025-06-15
chore: test conditional installs to speed up CI

SwenSchaeferjohann:
1. use robust caching. isolated per workflow
2. reduces to 20min CI time (if cache found)
3. we now run CI builds and tests for both JS v1 and v2 endpoints. gated by featureFlag. Note that _both_ JS versions interact only with the V2 programs, V2 indexer, and V2 prover.
4. splits JS-related CI into sdk/cli and V1/V2
4. adds `skip` option for `install.sh` because CI does not need local install of `Redis`. It uses Docker already.
5. more robust ci and `devenv.sh`


scripts/INSTALL.md
SwenSchaeferjohann, line 1: todo: remove

SwenSchaeferjohann:
closing in favor of #1822

PR #1822, opened 2025-06-16
chore: reduce ci times for non-js workflows

SwenSchaeferjohann:
reduce ci tests by parallelizing: 
- system-programs (system-cpi-test-v2, ["cargo-test-sbf -p system-cpi-v2-test"]) 
(40mins)
- Test concurrent-merkle-tree (40mins)
- Test program-libs (25)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,8 +28,6 @@ jobs:
 
       - name: Setup and build
         uses: ./.github/actions/setup-and-build
-        with:
-          skip-components: "redis,go,keys"
 
       - name: Run linters
         run: |


.github/workflows/light-system-programs-tests.yml
```diff
@@ -75,6 +79,8 @@ jobs:
 
       - name: Setup and build
         uses: ./.github/actions/setup-and-build
+        with:
+          skip-components: "redis"
```
ananas-block, line 83: I am surprised that this works, we use forester provingkeys in some of these tests and afaik forester provingkeys require redis

ananas-block (approved):

PR #1833, opened 2025-06-20
feat: implement batch ops cache with configurable TTL for batch ops

sergeytimoshin:


forester/src/cli.rs
```diff
@@ -66,6 +66,14 @@ pub struct StartArgs {
     )]
     pub transaction_max_concurrent_batches: usize,
 
+    #[arg(
+        long,
+        env = "FORESTER_TX_CACHE_TTL_SECONDS",
+        default_value = "180",
+        help = "Transaction cache TTL in seconds to prevent duplicate tx processing"
```
SwenSchaeferjohann: lets rename description for v2 to reflect the change in logic

PR #1838, opened 2025-06-21
refactor: batch ops

sergeytimoshin:
## Summary by CodeRabbit

* **New Features**
  * Added comprehensive end-to-end integration tests for both legacy and V2 pipelines, improving test coverage and validation of Merkle tree rollovers, queue processing, and registration logic.
  * Introduced Python scripts for analyzing and comparing performance metrics and TPS statistics from log files.
  * Added public functions for state and address Merkle tree rollovers.

* **Enhancements**
  * Refactored batch processing to use asynchronous streaming, enabling more efficient, incremental, and concurrent transaction handling.
  * Improved error handling and logging for queue metrics and processing rates.
  * Increased default batch instructions per transaction and RPC pool size for improved throughput.
  * Added support for API keys in client and connection configurations for enhanced security and flexibility.
  * Updated client configurations to include indexer URLs and API keys for improved integration.
  * Enhanced Forester status fetching to use indexer URL and API key.

* **Bug Fixes**
  * Fixed queue metric logging to provide clearer insights into queue states.

* **Documentation**
  * Expanded README with detailed environment variable documentation for running e2e V2 tests.

* **Chores**
  * Updated dependencies and script hashes for Photon indexer installation.
  * Simplified and consolidated GitHub Actions test workflows.
  * Removed deprecated traits and error types across the codebase.

* **Tests**
  * Added new utility functions and removed ignored attributes to enable more robust automated test execution.
  * Improved test utilities for slot and protocol config management.
  * Added legacy priority fee tests to validate fee capping and RPC fee estimation.

* **Refactor**
  * Removed deprecated traits and error types, simplifying codebase and generic constraints.
  * Streamlined batch and transaction processing logic for maintainability and performance.
  * Consolidated streaming abstractions for batch instruction generation and transaction submission.
  * Simplified trait bounds and removed `IndexerType` trait usage throughout the codebase.


forester/tests/e2e_v2_test.rs
```diff
@@ -0,0 +1,1372 @@
...
+            (
+                initial_next_index,
+                initial_sequence_number,
+                merkle_tree.get_root().unwrap(),
+            )
+        }
+    }
+}
+
+async fn verify_root_changed(
```
ananas-block, line 565: assert is too permissive imo.
Suggestion:
- assert next index equals an expected next index instead of root, this way we assert the exact number of updates

ananas-block: Additional suggestion:
- assert that all queue batches that can be have been inserted, to assert that the indexer is processing all trees

forester-utils/src/instructions/state_batch_append.rs
```diff
@@ -15,196 +17,172 @@ use light_prover_client::{
...
 
     if leaves_hash_chains.is_empty() {
-        trace!("No hash chains to process");
-        return Ok(Vec::new());
+        trace!("No hash chains to process, returning empty stream.");
+        return Ok((Box::pin(futures::stream::empty()), zkp_batch_size));
     }
 
-    wait_for_indexer(rpc, indexer).await?;
+    wait_for_indexer(&*rpc, &*indexer_guard).await?;
```
ananas-block: we can wait for all of these in parallel

forester-utils/src/instructions/state_batch_append.rs
```diff
@@ -15,196 +17,172 @@ use light_prover_client::{
...
-                trace!("Successfully generated proof for batch {}", i);
-                instruction_data_vec.push(InstructionDataBatchAppendInputs {
-                    new_root,
-                    compressed_proof: proof,
-                });
-            }
-            Err(e) => {
-                error!("Failed to generate proof for batch {}: {:?}", i, e);
-                return Err(e);
+        let proof_results = future::join_all(proof_futures).await;
```
ananas-block: not sure it's necessary for this pr, but could be faster if we don't wait for all proofs but just for the first batch.

ananas-block: or we just request a batch in each function call

ananas-block: Or maybe I got it wrong lmk

forester/src/processor/v2/common.rs
```diff
@@ -32,6 +34,136 @@ pub struct BatchContext<R: Rpc, I: Indexer> {
...
+    let mut instruction_buffer: Vec<D> = Vec::new();
+    let mut total_instructions_processed = 0;
+    let mut transactions_sent = 0;
+
+    while let Some(result) = instruction_stream.next().await {
+        let data = result?;
+        instruction_buffer.push(data);
+        total_instructions_processed += 1;
+
+        if instruction_buffer.len() >= context.ixs_per_tx {
```
ananas-block: Since we need to execute proofs strictly in sequence.
What do we do if we requested 10 proofs, `ixs_per_tx` is 5, and the 4th proof generation failed?

ananas-block (commented):
General flow:
1. separate thread for every tree
2. while in forester slot
  do: dispatch_tree_processing
         0. check whether a batch is available
           (we need to fetch accounts here and refetch in step 1,
            verify_batch_ready(), then again in get_append_instruction_stream())
         1. get queue and tree accounts once 
         2. get proofs
         3. stream proofs and send tx asap

Did I get that right?

- general feedback:
   1. code looks nice!
   2. we could try to remove some nesting layers there are multiple functions which wrap other functions in 5 lines and are only used once eg:
      - fn process_batched_operations<
      - async fn process_v2
      - async fn dispatch_tree_processing
   3. we need to fetch accounts here and refetch in step 1, verify_batch_ready(), then again in get_append_instruction_stream()

What do you think about suffixing functions with a number (nesting depth)?
(Sounds a bit stupid I know but could help to navigate the code.)


Prettier flow, could add it as doc comment somewhere:
```
// 1. Separate async task per tree
  tokio::spawn(async move {
      // 2. While in active phase and forester has slots
      for assigned_light_slot in tree_schedule.slots {
          // Inner loop every 25ms during slot window
          loop {
              // 3. Check eligibility and batch readiness
              let ready_state = verify_batch_ready(); // FETCH ACCOUNTS

              match ready_state {
                  ReadyForAppend => {
                      // 4. Create instruction stream
                      let (stream, _) = get_append_instruction_stream(); // FETCH ACCOUNTS AGAIN

                      // 5. Process stream - proofs generated concurrently
                      while let Some(instruction_data) = stream.next().await {
                          // 6. Buffer instructions and send tx when buffer full
                          if buffer.len() >= ixs_per_tx {
                              send_transaction_batch();
                              buffer.clear();
                          }
                      }
                  }
                  ReadyForNullify => { /* similar flow */ }
                  NotReady => { /* skip */ }
              }

              sleep(25ms); // Rate limiting
          }
      }
  });
```

forester/src/processor/v2/common.rs
sergeytimoshin: In this case we will send first N sucessfully generated proofs and after that will try to regenerated next ones

PR #1844, opened 2025-06-28
feat: opt ctoken

ananas-block:


programs/compressed-token/src/process_create_compressed_mint.rs
```diff
@@ -0,0 +1,270 @@
...
+        address_merkle_tree_root_index,
+        proof,
+    )?;
+
+    // 3. CPI to light-system-program
+    execute_cpi_invoke(&ctx, inputs_struct)
+}
+
+#[cfg(test)]
+mod tests {
```
ananas-block, line 171: move to separate test file

PR #1846, opened 2025-06-30
feat: add zero copy double iter

ananas-block:


program-libs/zero-copy/src/cyclic_vec.rs
```diff
@@ -417,3 +444,88 @@ fn test_private_getters() {
         assert_eq!(zcv.get_len_mut(), &mut (i + 1));
     }
 }
+
+#[test]
+fn test_double_ended_iterator() {
```
ananas-block, line 449: move tests to test dir

PR #1848, opened 2025-07-01
refactor: light-sdks detach account metas from account infos

ananas-block:


sdk-libs/sdk-pinocchio/src/error.rs
```diff
@@ -68,12 +69,20 @@ pub enum LightSdkError {
     MetaCloseInputIsNone,
     #[error("CPI accounts index out of bounds: {0}")]
     CpiAccountsIndexOutOfBounds(usize),
+    #[error("Invalid CPI context account")]
+    InvalidCpiContextAccount,
+    #[error("Invalid sol pool pda account")]
+    InvalidSolPoolPdaAccount,
+    #[error("CpigAccounts accounts slice starts with an invalid account. It should start with LightSystemProgram SySTEM1eSU2p4BGQfQpimFEWWSC1XDFeun3Nqzz3rT7.")]
```
SwenSchaeferjohann:
```suggestion
    #[error("CpiAccounts slice starts with an invalid account. It should start with LightSystemProgram SySTEM1eSU2p4BGQfQpimFEWWSC1XDFeun3Nqzz3rT7.")]
```

PR #1850, opened 2025-07-04
feat: compressible accounts

SwenSchaeferjohann:
adds low level sdk helpers

- compress_pda
- compress_pda_new
- compress_multiple_pdas_new
- decompress_idempotent
- decompress_multiple_idempotent

caller programs can define
- which PDAs to decompress
- slots after which a PDA should be compressible (`slots_until_compression`)
- `rent_payer` (anyone)
- `rent_recipient` (unchecked, hc)
-  compressed-accounts address must be part of the onchain PDA's seeds


sdk-libs/sdk/src/compressible/compress_pda.rs
```diff
@@ -0,0 +1,120 @@
...
+// - check if any explicit checks required for compressed account?
+// - consider multiple accounts per ix.
+pub fn compress_pda<A>(
+    pda_account: &AccountInfo,
+    compressed_account_meta: &CompressedAccountMeta,
+    proof: ValidityProof,
+    cpi_accounts: CpiAccounts,
+    owner_program: &Pubkey,
+    rent_recipient: &AccountInfo,
+) -> Result<(), LightSdkError>
```
ananas-block, line 55: imo it makes sense to use an input struct

program-tests/sdk-test/src/create_dynamic_pda.rs
```diff
@@ -0,0 +1,70 @@
...
+    error::LightSdkError,
+    instruction::{PackedAddressTreeInfo, ValidityProof},
+};
+use light_sdk_types::CpiAccountsConfig;
+use solana_program::account_info::AccountInfo;
+use solana_program::pubkey::Pubkey;
+
+use crate::decompress_dynamic_pda::MyPdaAccount;
+
+pub const ADDRESS_SPACE: Pubkey = pubkey!("CLEuMG7pzJX9xAuKCFzBP154uiG1GaNo4Fq7x6KAcAfG");
```
ananas-block, line 15: the address tree pubkey should be different

PR #1857, opened 2025-07-07
feat: `compressible` proc macro draft

SwenSchaeferjohann:


program-tests/sdk-test/src/compress_dynamic_pda.rs
```diff
@@ -0,0 +1,59 @@
...
+    let instruction_data = CompressFromPdaInstructionData::deserialize(&mut instruction_data)
+        .map_err(|_| LightSdkError::Borsh)?;
+
+    let pda_account = &accounts[1];
+
+    // CHECK: hardcoded rent recipient.
+    let rent_recipient = &accounts[2];
+    if rent_recipient.key != &crate::create_dynamic_pda::RENT_RECIPIENT {
+        return Err(LightSdkError::ConstraintViolation);
+    }
```
ananas-block: no need to check the recipient if it is part of the account.
We can trust data from Solana and compressed accounts.

sdk-libs/sdk/src/compressible/compress_pda_new.rs
```diff
@@ -0,0 +1,199 @@
...
+    let dest_starting_lamports = rent_recipient.lamports();
+    **rent_recipient.try_borrow_mut_lamports()? = dest_starting_lamports
+        .checked_add(total_lamports)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+
+    for pda_account in pda_accounts {
+        // Decrement source account lamports
+        **pda_account.try_borrow_mut_lamports()? = 0;
+        // Clear all account data
+        pda_account.try_borrow_mut_data()?.fill(0);
```
ananas-block: this is not necessary if you set lamports to zero, but can cost a lot of CU

sdk-libs/sdk/src/compressible/compress_pda_new.rs
```diff
@@ -0,0 +1,199 @@
...
+        // Check that the PDA account is owned by the caller program
+        if pda_account.owner != owner_program {
+            msg!(
+                "Invalid PDA owner for {}. Expected: {}. Found: {}.",
+                pda_account.key,
+                owner_program,
+                pda_account.owner
+            );
+            return Err(LightSdkError::ConstraintViolation);
+        }
```
ananas-block: The owner check is already enforced by the runtime.

sdk-libs/sdk/src/compressible/compress_pda_new.rs
```diff
@@ -0,0 +1,199 @@
...
+            .get_tree_account_info(params.address_merkle_tree_account_index as usize)?;
+        if address_tree_account.pubkey() != *expected_address_space {
+            msg!(
+                "Invalid address space. Expected: {}. Found: {}.",
+                expected_address_space,
+                address_tree_account.pubkey()
+            );
+            return Err(LightSdkError::ConstraintViolation);
+        }
+    }
```
ananas-block: instead of checking it we could also hardcode it as a constant

sdk-libs/sdk/src/compressible/compress_pda_new.rs
```diff
@@ -0,0 +1,199 @@
...
+        + Default
+        + CompressionTiming
+        + Clone,
+{
+    if pda_accounts.len() != addresses.len()
+        || pda_accounts.len() != new_address_params.len()
+        || pda_accounts.len() != output_state_tree_indices.len()
+    {
+        return Err(LightSdkError::ConstraintViolation);
+    }
```
ananas-block: It's probably more dev friendly to bundle these into a struct so that it's not possible to create len mismatches.

```
struct CompressingPda {
    account_info: AccountInfo,
    address_params: PackedNewAddressParams,
    out_tree_index: u8,
}
```

sdk-libs/sdk/src/compressible/compress_pda_new.rs
```diff
@@ -0,0 +1,199 @@
...
+/// * `Err(LightSdkError)` if there was an error
+pub fn compress_multiple_pdas_new<'info, A>(
+    pda_accounts: &[&AccountInfo<'info>],
+    addresses: &[[u8; 32]],
+    new_address_params: Vec<PackedNewAddressParams>,
+    output_state_tree_indices: &[u8],
+    proof: ValidityProof,
+    cpi_accounts: CpiAccounts<'_, 'info>,
+    owner_program: &Pubkey,
+    rent_recipient: &AccountInfo<'info>,
```
ananas-block: it's probably ok if its just a Pubkey. With Solana and anchor entrypoint, it safes CU to pass fewer AccountInfos and use instruction data instead. We can zero copy instruction data but not AccountInfos.

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,199 @@
...
+/// * `Ok(())` if all compressed accounts were decompressed successfully or PDAs already exist
+/// * `Err(LightSdkError)` if there was an error
+pub fn decompress_multiple_idempotent<'info, A>(
+    pda_accounts: &[&AccountInfo<'info>],
+    compressed_accounts: Vec<LightAccount<'_, A>>,
+    proof: ValidityProof,
+    cpi_accounts: CpiAccounts<'_, 'info>,
+    owner_program: &Pubkey,
+    rent_payer: &AccountInfo<'info>,
+    system_program: &AccountInfo<'info>,
```
ananas-block: if you pass cpi_accounts, there is no need to pass the system program separately.

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,199 @@
...
+/// * `owner_program` - The program that will own the PDAs
+/// * `rent_payer` - The account to pay for PDA rent
+/// * `system_program` - The system program
+///
+/// # Returns
+/// * `Ok(())` if all compressed accounts were decompressed successfully or PDAs already exist
+/// * `Err(LightSdkError)` if there was an error
+pub fn decompress_multiple_idempotent<'info, A>(
+    pda_accounts: &[&AccountInfo<'info>],
+    compressed_accounts: Vec<LightAccount<'_, A>>,
```
ananas-block:
```suggestion
    compressed_accounts: &[LightAccount<'_, A>],
```

We should avoid vectors.

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,199 @@
...
+            rent_minimum_balance,
+            space as u64,
+            owner_program,
+        );
+
+        // Add bump to seeds for signing
+        let bump_seed = [pda_bump];
+        let mut signer_seeds = seeds.clone();
+        signer_seeds.push(&bump_seed);
+        let signer_seeds_refs: Vec<&[u8]> = signer_seeds.iter().map(|s| *s).collect();
```
ananas-block: this is really inefficient. There is a max number of seeds solana accepts, I think it is 16.
We can use an ArrayVec with max size 16 to avoid any heap operations here.

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        // Get the compressed account address
+        let compressed_address = compressed_account
+            .address()
+            .ok_or(LightSdkError::ConstraintViolation)?;
+
+        // Derive onchain PDA using the compressed address as seed
+        let seeds: Vec<&[u8]> = vec![&compressed_address];
+
+        let (pda_pubkey, pda_bump) = Pubkey::find_program_address(&seeds, owner_program);
```
ananas-block: we should use `create_program_address` and pass the bump

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,199 @@
...
+            continue;
+        }
+
+        // Get the compressed account address
+        let compressed_address = compressed_account
+            .address()
+            .ok_or(LightSdkError::ConstraintViolation)?;
+
+        // Derive onchain PDA using the compressed address as seed
+        let seeds: Vec<&[u8]> = vec![&compressed_address];
```
ananas-block:
```suggestion
        let seeds: [&[u8];1] = [&compressed_address];
```

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,199 @@
...
+        + Clone
+        + CompressionTiming,
+{
+    // Get current slot and rent once for all accounts
+    let clock = Clock::get().map_err(|_| LightSdkError::Borsh)?;
+    let current_slot = clock.slot;
+    let rent = Rent::get().map_err(|_| LightSdkError::Borsh)?;
+
+    // Calculate space needed for PDA (same for all accounts of type A)
+    let space = std::mem::size_of::<A>() + 8; // +8 for discriminator
```
ananas-block: this is not reliable if the account struct contains vectors or options. We need to pass size as an input.

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,199 @@
...
+        let discriminator = A::LIGHT_DISCRIMINATOR;
+        pda_account.try_borrow_mut_data()?[..8].copy_from_slice(&discriminator);
+
+        // Write data to PDA
+        decompressed_pda
+            .serialize(&mut &mut pda_account.try_borrow_mut_data()?[8..])
+            .map_err(|_| LightSdkError::Borsh)?;
+
+        // Zero the compressed account
+        compressed_account.account = A::default();
```
ananas-block: this doesn't properly zero the compressed account.
Option 1:
- create a new `LightAccount` struct which only holds the address and no data.
This will add a lot of zeroes to the light system program cpi.

Option 2:
- add `zero_out` or `remove_data` method to `LightAccount` which removes the data.
- maybe we need to add and set a bool flag in `LightAccount` 


Option 2 is probably better because we should keep the discriminator attached to the address.

In either case the clients need to handle this case that they cannot deserialize the empty account struct from the fetched compressed account because photon will not return any zero bytes if we don't send them in the light system program cpi.

sdk-libs/macros/Cargo.toml
```diff
@@ -11,7 +11,7 @@ proc-macro2 = { workspace = true }
 quote = { workspace = true }
 syn = { workspace = true }
 solana-pubkey = { workspace = true, features = ["curve25519", "sha2"] }
-
+heck = "0.4.1"
```
ananas-block, line 18: what da heck

program-tests/sdk-test/src/decompress_dynamic_pda.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub meta: CompressedAccountMeta,
+    pub data: MyPdaAccount,
+}
+
+#[derive(
+    Clone, Debug, Default, LightHasher, LightDiscriminator, BorshDeserialize, BorshSerialize,
+)]
+pub struct MyPdaAccount {
+    pub last_written_slot: u64,
+    pub compression_delay: u64,
```
ananas-block: we should provide a header struct.
```
struct Header {
    pub discriminator: [u8;8], // debatable
    pub rent_recipient: [u8;32],
    pub last_written_slot: u64,
    pub compression_delay: u64, // can make this a u32 because we can just add it on `last_written_slot`
    pub updatable: bool,
    pub compress_sol: bool,
    pub version: Enum(u16)
    _padding: [u8;4], // for memory alignment so that we can do a zero copy
}

Comment (ananas-block) [reply program-tests/sdk-test/src/decompress_dynamic_pda.rs]: the header should not be part of the compressed account, and should not derive `LightHasher`.

File: program-tests/sdk-test/src/decompress_dynamic_pda.rs [hunk 16]
```diff
@@ -0,0 +1,161 @@
...
+
+    // Get PDA accounts (after fixed accounts, before system accounts)
+    let pda_accounts_start = 3;
+    let pda_accounts_end = instruction_data.system_accounts_offset as usize;
+    let pda_accounts = &accounts[pda_accounts_start..pda_accounts_end];
+
+    // Set up CPI accounts
+    let mut config = CpiAccountsConfig::new(crate::LIGHT_CPI_SIGNER);
+    config.sol_pool_pda = false;
+    config.sol_compression_recipient = false;
```
ananas-block: new should already set this correctly

ananas-block: and you can also just do `CpiAccounts::new(crate::LIGHT_CPI_SIGNER)` to avoid the manual config creation.

ananas-block: Ah maybe this is why you experimented with it.
It would be really cool to compress the sol with the account, then we can pay the rent with the sol from the compressed account.

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
SwenSchaeferjohann: added `.size()` to lightaccount - since we already pass the data there

sdk-libs/sdk/src/compressible/compress_pda.rs
```diff
@@ -0,0 +1,114 @@
...
+{
+    // Check that the PDA account is owned by the caller program
+    if pda_account.owner != owner_program {
+        msg!(
+            "Invalid PDA owner. Expected: {}. Found: {}.",
+            owner_program,
+            pda_account.owner
+        );
+        return Err(LightSdkError::ConstraintViolation);
+    }
```
ananas-block: we can use the light-account-check crate instead of manual checks

sdk-libs/sdk/src/compressible/compress_pda.rs
```diff
@@ -0,0 +1,114 @@
...
+    cpi_inputs.invoke_light_system_program(cpi_accounts)?;
+
+    // Close the PDA account
+    // 1. Transfer all lamports to the rent recipient
+    let dest_starting_lamports = rent_recipient.lamports();
+    **rent_recipient.try_borrow_mut_lamports()? = dest_starting_lamports
+        .checked_add(pda_lamports)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+    // 2. Decrement source account lamports
+    **pda_account.try_borrow_mut_lamports()? = 0;
```
ananas-block: I will need this logic as well in the ctoken program we should create a shared function which closes an account this way

sdk-libs/client/src/compressible.rs
```diff
@@ -0,0 +1,421 @@
+use solana_instruction::{AccountMeta, Instruction};
+use solana_pubkey::Pubkey;
+
+#[cfg(feature = "anchor")]
+use anchor_lang::{AnchorDeserialize, AnchorSerialize};
+#[cfg(not(feature = "anchor"))]
+use borsh::{BorshDeserialize as AnchorDeserialize, BorshSerialize as AnchorSerialize};
```
ananas-block: this should not be in `light-client`, it should either be in `light-sdk` since this only creates instructions or be it's own crate.

sdk-libs/program-test/src/program_test/light_program_test.rs
```diff
@@ -75,7 +75,7 @@ impl LightProgramTest {
             if context.config.skip_startup_logs {
                 context.config.no_logs = true;
             }
-            initialize_accounts(&mut context, &config, &keypairs).await?;
+            context.test_accounts = initialize_accounts(&mut context, &config, &keypairs).await?;
```
ananas-block: the accounts are already set before with `get_program_test_test_accounts` we should make sure that those contain the new accounts as well.

program-tests/utils/src/test_keypairs.rs
```diff
@@ -64,10 +64,15 @@ pub fn from_target_folder() -> TestKeypairs {
...
         group_pda_seed: Keypair::new(),
+        batched_state_merkle_tree_2: Keypair::from_bytes(&BATCHED_STATE_MERKLE_TREE_TEST_KEYPAIR_2)
+            .unwrap(),
+        batched_output_queue_2: Keypair::from_bytes(&BATCHED_OUTPUT_QUEUE_TEST_KEYPAIR_2).unwrap(),
+        batched_cpi_context_2: Keypair::from_bytes(&BATCHED_CPI_CONTEXT_TEST_KEYPAIR_2).unwrap(),
     }
 }
 
 pub fn for_regenerate_accounts() -> TestKeypairs {
+    // Note: this requries your machine to have the light-keypairs dir with the correct keypairs.
```
ananas-block, line 75: we should change that, the same way I am changing the address for the address Merkle tree in light-program-test setup.

xtask/src/create_batch_state_tree.rs
```diff
@@ -30,6 +30,68 @@ pub struct Options {
     config: Option<String>,
 }
 
+pub async fn write_account_state_to_file(options: Options) -> anyhow::Result<()> {
```
ananas-block: this is probably not necessary since you added the second tree creation to `light-program-test` let's remove

ananas-block (commented):
Ideas to split this up:
1. adding the second batched Merkle tree could easily be a separate pr

program-tests/sdk-test/src/create_dynamic_pda.rs
```diff
@@ -0,0 +1,143 @@
...
+    let config = CompressibleConfig::load_checked(config_account, &crate::ID)?;
+
+    // CHECK: rent recipient from config
+    if rent_recipient.key != &config.rent_recipient {
+        solana_program::msg!(
+            "rent recipient mismatch {:?} != {:?}",
+            rent_recipient.key,
+            config.rent_recipient
+        );
+        return Err(LightSdkError::ConstraintViolation);
```
ananas-block: more specific errors make debugging easier.

program-tests/sdk-test/src/update_config.rs
```diff
@@ -0,0 +1,46 @@
...
+    let mut instruction_data = instruction_data;
+    let instruction_data = UpdateConfigInstructionData::deserialize(&mut instruction_data)
+        .map_err(|_| LightSdkError::Borsh)?;
+
+    // Get accounts
+    let config_account = &accounts[0];
+    let authority = &accounts[1];
+
+    // Verify the config PDA
+    let (expected_pda, _) = CompressibleConfig::derive_pda(&crate::ID);
```
ananas-block: we only allow one config per program, what if we want different configs for different account types?

ananas-block: Maybe we should store a vector of configs.

sdk-libs/sdk/src/compressible/config.rs
```diff
@@ -0,0 +1,460 @@
...
+const BPF_LOADER_UPGRADEABLE_ID: Pubkey =
+    Pubkey::from_str_const("BPFLoaderUpgradeab1e11111111111111111111111");
+
+/// Global configuration for compressible accounts
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, LightDiscriminator)]
+pub struct CompressibleConfig {
+    /// Config version for future upgrades
+    pub version: u8,
+    /// Discriminator for account validation
+    pub discriminator: [u8; 8],
```
ananas-block: discriminator of the config account itself or the account the config is meant for?
1. `config account itself`
    - atypical to put another field before the discriminator.

sdk-libs/sdk/src/compressible/config.rs
```diff
@@ -0,0 +1,460 @@
...
+    /// Number of slots to wait before compression is allowed
+    pub compression_delay: u32,
+    /// Authority that can update the config
+    pub update_authority: Pubkey,
+    /// Account that receives rent from compressed PDAs
+    pub rent_recipient: Pubkey,
+    /// Address space for compressed accounts (exactly 1 address_tree allowed)
+    pub address_space: Vec<Pubkey>,
+    /// PDA bump seed
+    pub bump: u8,
```
ananas-block, line 32: static values should be before vectors.

sdk-libs/sdk/src/compressible/config.rs
```diff
@@ -0,0 +1,460 @@
+use std::collections::HashSet;
+
+#[cfg(feature = "anchor")]
+use anchor_lang::{AnchorDeserialize as BorshDeserialize, AnchorSerialize as BorshSerialize};
+#[cfg(not(feature = "anchor"))]
+use borsh::{BorshDeserialize, BorshSerialize};
```
ananas-block: use crate::{AnchorDeserialize, AnchorSerialize}; instead
in lib.rs we assign it based on the anchor feature in a unified way

sdk-libs/sdk/src/compressible/config.rs
```diff
@@ -0,0 +1,460 @@
...
+pub const COMPRESSIBLE_CONFIG_SEED: &[u8] = b"compressible_config";
+pub const MAX_ADDRESS_TREES_PER_SPACE: usize = 1;
+
+
+const BPF_LOADER_UPGRADEABLE_ID: Pubkey =
+    Pubkey::from_str_const("BPFLoaderUpgradeab1e11111111111111111111111");
+
+/// Global configuration for compressible accounts
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, LightDiscriminator)]
+pub struct CompressibleConfig {
```
ananas-block, line 20: We should make this usable with the anchor `Account` abstraction.

sdk-libs/sdk/src/compressible/config.rs
```diff
@@ -0,0 +1,460 @@
...
+/// # Required Validation (must be done by caller)
+/// The caller MUST validate that the signer is the program's upgrade authority
+/// by checking against the program data account. This cannot be done in the SDK
+/// due to dependency constraints.
+///
+/// # Returns
+/// * `Ok(())` if config was created successfully
+/// * `Err(ProgramError)` if there was an error
+#[allow(clippy::too_many_arguments)]
+pub fn process_initialize_compression_config_unchecked<'info>(
```
ananas-block:
```suggestion
pub fn process_initialize_compression_config_account_info<'info>(
```

As far as I can tell we perform checks in this instruction -> unchecked is misleading.

sdk-libs/sdk/src/compressible/compress_pda_new.rs
```diff
@@ -0,0 +1,373 @@
...
+    info: &mut AccountInfo<'info>,
+    sol_destination: AccountInfo<'info>,
+) -> Result<()> {
+    // Transfer all lamports from the account to the destination
+    let lamports_to_transfer = info.lamports();
+
+    // Use try_borrow_mut_lamports for proper borrow management
+    **info
+        .try_borrow_mut_lamports()
+        .map_err(|_| LightSdkError::ConstraintViolation)? = 0;
```
ananas-block: meaningful errors instead of `ConstraintViolation`

sdk-libs/sdk/src/account.rs
```diff
@@ -230,18 +281,57 @@ impl<'a, A: AnchorSerialize + AnchorDeserialize + LightDiscriminator + DataHashe
...
-            output.data_hash = self.account.hash::<Poseidon>()?;
-            output.data = self
-                .account
-                .try_to_vec()
-                .map_err(|_| LightSdkError::Borsh)?;
+            if self.empty_data {
+                // TODO: review security.
+                output.data = Vec::new();
+                output.output_merkle_tree_index = 1;
+                output.data_hash = [1; 32];
```
ananas-block:
```suggestion
                output.data_hash = [1; 32];
```

ananas-block: data is already an empty vec, and we shouldn't arbitrarily overwrite the output_merkle_tree_index.

sdk-libs/sdk/src/compressible/config.rs
SwenSchaeferjohann: I don't actually have use for it at all, will remove

sdk-libs/sdk/src/account.rs
```diff
@@ -230,18 +281,57 @@ impl<'a, A: AnchorSerialize + AnchorDeserialize + LightDiscriminator + DataHashe
...
+    /// Calculate the size required for this account when stored on-chain.
+    /// This includes the 8-byte discriminator plus the serialized data size.
+    pub fn size(&self) -> Result<usize, LightSdkError> {
+        let data_size = self
+            .account
+            .try_to_vec()
+            .map_err(|_| LightSdkError::Borsh)?
+            .len();
+        Ok(8 + data_size) // 8 bytes for discriminator + data size
+    }
```
ananas-block: this is super expensive, we should do it another way.

sdk-libs/sdk/src/account.rs
```diff
@@ -230,18 +281,57 @@ impl<'a, A: AnchorSerialize + AnchorDeserialize + LightDiscriminator + DataHashe
...
+    ) -> Result<&'info solana_pubkey::Pubkey, LightSdkError> {
+        let merkle_tree_index = self
+            .in_account_info()
+            .as_ref()
+            .ok_or(LightSdkError::ConstraintViolation)?
+            .merkle_context
+            .merkle_tree_pubkey_index as usize;
+
+        Ok(cpi_accounts.get_tree_account_info(merkle_tree_index)?.key)
+    }
```
ananas-block: potentially dangerous because we need to discourage usage patterns where in and out Merkle tree can be accidentially hardcoded to be the same tree. (Accounts will get stuck in one tree and be unusable once the tree is full.)

sdk-libs/sdk/src/account.rs
```diff
@@ -155,6 +158,53 @@ impl<'a, A: AnchorSerialize + AnchorDeserialize + LightDiscriminator + DataHashe
...
+    /// This is used when compressing a PDA - we know the compressed account exists
+    /// but is empty (data: [], data_hash: [1; 32]).
+    pub fn new_mut_without_data(
+        owner: &'a Pubkey,
+        input_account_meta: &impl CompressedAccountMetaTrait,
+    ) -> Result<Self, LightSdkError> {
+        let input_account_info = {
+            let tree_info = input_account_meta.get_tree_info();
+            InAccountInfo {
+                data_hash: [1; 32], // TODO: review security.
```
ananas-block: we should make this a constant

sdk-libs/sdk/Cargo.toml
```diff
@@ -28,13 +27,19 @@ solana-msg = { workspace = true }
...
 solana-program-error = { workspace = true }
 solana-instruction = { workspace = true }
+solana-system-interface = { workspace = true }
+solana-clock = { workspace = true }
+solana-sysvar = { workspace = true }
+solana-rent = { workspace = true }
 
-anchor-lang = { workspace = true, optional = true }
+
+anchor-lang = { workspace = true }
```
ananas-block:
```suggestion
anchor-lang = { workspace = true, optional = true }
```

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,172 @@
...
+        )?;
+
+        // Initialize PDA with decompressed data and update slot
+        let mut decompressed_pda = compressed_account.account.clone();
+
+        msg!(
+            "Initializing compression info for PDA {:?}",
+            pda_account.key
+        );
+        *decompressed_pda.compression_info_mut_opt() = Some(super::CompressionInfo::new()?);
```
ananas-block: shouldn't`CompressionInfo` be initialized here?

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,172 @@
...
+/// * `cpi_accounts` - Accounts needed for CPI
+/// * `owner_program` - The program that will own the PDAs
+/// * `rent_payer` - The account to pay for PDA rent
+/// * `address_space` - The address space for the compressed accounts
+///
+/// # Returns
+/// * `Ok(Vec<CompressedAccountInfo>)` - CompressedAccountInfo for CPI batching
+/// * `Err(LightSdkError)` if there was an error
+pub fn prepare_accounts_for_decompress_idempotent<'info, T>(
+    pda_accounts: &[&AccountInfo<'info>],
```
ananas-block:
```suggestion
    solana_accounts: &[&AccountInfo<'info>],
```
would be more explicit, I got confused looking at `pda_account.data_is_empty() `

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,172 @@
...
+    compressed_accounts: Vec<LightAccount<'_, T>>,
+    signer_seeds: &[&[&[u8]]],
```
ananas-block: we need to use CPI_SIGNER_SEEDS right?
(If so hardcoding will save a lot of pain.)

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
```diff
@@ -0,0 +1,172 @@
...
+    cpi_accounts: &CpiAccounts<'_, 'info>,
+    owner_program: &Pubkey,
```
ananas-block: `owner_program` is already contained in `cpi_accounts` and `compressed_accounts`

ananas-block:
Main questions:
- where do we set compression info for solana pdas when decompressing?
- a single config is likely limiting (fixed derivation path limits it to one, version field probably doesn't help, could just make it a vector.)

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
SwenSchaeferjohann: at decompress the onchain account's compressioninfo is `None` (because its just been opened again) until we set it here.
we don't actually care what the state was until here - we can just do `new()` which sets it to the current_slot + `CompressionState.Decompressed`

SwenSchaeferjohann:
> * where do we set compression info for solana pdas when decompressing?

here https://github.com/Lightprotocol/light-protocol/pull/1857#discussion_r2234099553 (we can rename `new()` to `new_with_current_slot()` or `new_decompressed()` to make this more explicit) 

edit: renamed `new()` to `new_decompressed()`

> * a single config is likely limiting (fixed derivation path limits it to one, version field probably doesn't help, could just make it a vector.)

this or we could add a u8 config_index to the derivation path and have 1 account per config - what's better?

xtask/src/create_batch_state_tree.rs
SwenSchaeferjohann: removed

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
SwenSchaeferjohann: removed

sdk-libs/sdk/src/account.rs
SwenSchaeferjohann: not using it anymore anyway. deleted it

sdk-libs/sdk/src/account.rs
SwenSchaeferjohann: replaced with a compile-time `size()` trait that devs must implement for their compressible account struct

in anchor they can reuse the anchor tooling, so it's simple:

```rust
impl Size for UserRecord {
    fn size(&self) -> usize {
        Self::DISCRIMINATOR.len() + Self::INIT_SPACE
    }
}
```

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
SwenSchaeferjohann: renamed

sdk-libs/sdk/src/account.rs
SwenSchaeferjohann: fixed

sdk-libs/sdk/src/compressible/decompress_idempotent.rs
SwenSchaeferjohann: those are just the signer_seeds for the create_account cpi to the solana systemprogram

our light cpi signer seeds are via invoke_light_system_program

sdk-libs/sdk/src/compressible/config.rs
SwenSchaeferjohann: the reason for having the config be fully native is so it's fully usable by both native programs and anchor programs. everything config related is also end-to-end static; so no extra checks needed by dev

ananas-block: one doesn't prevent the other. If anchor devs should use this account outside of an autogenerated instruction, it will be more in line with anchor programming mode, they will feel much safer if they can use the anchor Account abstraction.

PR #1912, opened 2025-08-17
feat: add API key support to prover client

sergeytimoshin:


forester/src/processor/v2/address.rs
```diff
@@ -27,7 +27,8 @@ where
     let config = AddressUpdateConfig {
         rpc_pool: ctx.rpc_pool.clone(),
         merkle_tree_pubkey: ctx.merkle_tree,
-        prover_url: ctx.prover_url.clone(),
+        prover_url: construct_prover_url(&ctx.prover_url, "/address-append"),
```
SwenSchaeferjohann: this is supposed to have the endpoint hardcoded here? @sergeytimoshin

SwenSchaeferjohann (approved):

forester/src/processor/v2/address.rs
sergeytimoshin: You are right, I'll clean it up.

PR #1941, opened 2025-09-22
refactor: forester cleanup 1 

sergeytimoshin:
- switched from queue offsets to indices for cleaner queue handling
- fixed registration phase issues - forester now properly skips epochs it can't register for
- improved transaction processing with active phase checks
- cleaned up verbose logging (removed TPS metrics spam)
- added support for processing specific trees via CLI


SwenSchaeferjohann:
do we have notifs for when epoch registration fails

sergeytimoshin:
> do we have notifs for when epoch registration fails

not yet, good idea.

sergeytimoshin:
closing this for https://github.com/Lightprotocol/light-protocol/pull/1942

PR #1942, opened 2025-09-23
feat: gnark14, forester

sergeytimoshin:
forester: 
- switched from queue offsets to indices for cleaner queue handling
- fixed registration phase issues - forester now properly skips epochs it can't register for
- improved transaction processing with active phase checks
- cleaned up verbose logging (removed TPS metrics spam)
- added support for processing specific trees via CLI

gnark:
- gnark updated to 0.14
- gnark-crypto updated to 0.19
- updated formal verification from GatesGnark8 to GatesGnark12
- regenerated all proving and verifying keys for v0.14.0 compatibility
- prover code restructured, legacy circuits => v1, merkle circuits with PublicInputHash + batch circuits => v2
- fixed key generation and download scripts for new key format


forester-utils/src/instructions/address_batch_update.rs
```diff
@@ -21,9 +21,10 @@ use light_prover_client::{
 use light_sparse_merkle_tree::SparseMerkleTree;
 use tracing::{debug, error, info, warn};
 
-use crate::{error::ForesterUtilsError, rpc_pool::SolanaRpcPool};
+use crate::{error::ForesterUtilsError, rpc_pool::SolanaRpcPool, utils::wait_for_indexer};
 
 const MAX_PHOTON_ELEMENTS_PER_CALL: usize = 500;
+const MAX_PROOFS_PER_TX: usize = 3;
```
ananas-block, line 27: can't we do more than 3 proofs?

sergeytimoshin: no, unfortunately

sdk-tests/client-test/tests/light_client.rs
```diff
@@ -742,6 +742,9 @@ async fn create_address(
     owner: Pubkey,
     merkle_tree: Pubkey,
 ) -> Result<([u8; 32], Signature), RpcError> {
+    #[cfg(feature = "v2")]
+    let address_merkle_tree = rpc.get_address_tree_v2();
+    #[cfg(not(feature = "v2"))]
```
ananas-block, line 747:
```suggestion
```
V1 address trees should still be available with v2 flag. We will deprecate it gradually.

sdk-libs/client/src/indexer/photon_indexer.rs
```diff
@@ -1602,79 +1598,79 @@ impl Indexer for PhotonIndexer {
...
+                                        account_hash,
+                                    }
+                                })
+                                .collect();
+
+                            Ok(Response {
+                                context: Context {
+                                    slot: api_result.context.slot,
+                                },
+                                value: (proofs, Some(api_result.first_value_queue_index as u64)),
```
ananas-block: Can we return a result struct instead of the tuple?

forester-utils/src/instructions/address_batch_update.rs
ananas-block: 3 proofs are 384 bytes + 96 bytes public inputs (3 new roots) + 5 accounts 192 bytes + 1 signature 64 bytes + 3 * 12 bytes registry instruction overhead = `740` bytes a tx has space for 1232 bytes instruction data

One proof (128) + new root (32) = 160 bytes
[Batch nullify requires](https://github.com/Lightprotocol/light-protocol/blob/0b783c9fa16433cf3b40324dc1d3db5b9d0ff4cf/programs/account-compression/src/instructions/batch_nullify.rs#L15-L24) 4 accounts

[Batch append requires](https://github.com/Lightprotocol/light-protocol/blob/0b783c9fa16433cf3b40324dc1d3db5b9d0ff4cf/programs/account-compression/src/instructions/batch_append.rs#L15) 5 accounts

There are 12 additional bytes [per batch nullify instruction](https://github.com/Lightprotocol/light-protocol/blob/0b783c9fa16433cf3b40324dc1d3db5b9d0ff4cf/programs/registry/src/lib.rs#L513-L515):
1. 8 bytes anchor instruction discriminator of the registry program
2. vec size 4 bytes
3. bump 1 byte
(Its the same for batch nullify and batch append)

For 5 batch append instructions the transaction data should be roughly:
`5 * (128 + 32 + 12) + 5 * 32 + 64 = 1,084`

hm that should still work

PR #1946, opened 2025-09-27
feat: sdk v2 2.3

ananas-block:


program-libs/account-checks/src/packed_accounts.rs
```diff
@@ -0,0 +1,33 @@
+use std::panic::Location;
+
+use crate::{AccountError, AccountInfoTrait};
+
+/// Dynamic accounts slice for index-based access
+/// Contains mint, owner, delegate, merkle tree, and queue accounts
```
SwenSchaeferjohann: odd inline doc since this is generic no?

ananas-block: also this is a general primitive and has to relation to mints etc.

ananas-block: need to clean up inline docs in general

SwenSchaeferjohann (approved):

PR #2014, opened 2025-10-25
chore: pre release

ananas-block:
## Summary by CodeRabbit

* **New Features**
  * Introduced separate V1 and V2 address tree network fee constants (5000 and 10000 respectively).
  * Added authorization validation for batched merkle tree initialization.
  * Implemented version-aware input compressed account tracking.

* **Bug Fixes**
  * Improved error handling for missing legacy merkle context.

* **Refactor**
  * Updated default network fees and introduced distinct address network fee configuration.
  * Reorganized fee calculation logic into per-input and per-address components.
  * Deprecated program-owned tree creation paths; V1 address trees now loaded from persistent storage.
  * Simplified feature defaults across multiple programs.


program-tests/account-compression-test/tests/batched_merkle_tree_test.rs
```diff
@@ -1,5 +1,3 @@
-#![cfg(feature = "test-sbf")]
```
ananas-block, line 1: removed on purpose, these are not necessary since we dont run non sbf tests in the same test crate and prevent rust analyzer from compiling the file.

PR #2023, opened 2025-10-28
refactor: get program accounts

ananas-block:


SwenSchaeferjohann (approved):
lgtm. worth upstreaming @ananas-block ?

PR #2034, opened 2025-11-05
fix: add realloc to cpi context account reinit ix

ananas-block:
**Changes:**
- realloc cpi context account to `14020` bytes from 20kb
- add remaining cpi context account capacity assert to reinit cpi context account test

**Context:**
- at 20kb size cpi context accounts have `12,968` bytes remaining capacity for output account data
- these bytes are necessary to store compressed account data used in the cpi context
- `12,968` bytes is greater than the light system program -> account compression program cpi can hold which is a footgun
- with the realloc to `14020` bytes we reduce the remaining capacity to `6,500` that should work in most cases.
- cpi context can support up to `30` output accounts with max `6,500` bytes account data eg `216` bytes per compressed account, or a number of smaller and bigger compressed accounts
- The account data is stored to be exposed in the light system program -> account compression program cpi for the indexer.

## Summary by CodeRabbit

* **New Features**
  * Added account remaining-capacity reporting for clearer state visibility

* **Bug Fixes**
  * Verify ownership before account resize to prevent unintended data truncation; preserve pre-read state during reinitialization

* **Chores**
  * Switched CPI context account sizing to versioned constants (V1 deprecated, V2 active) and updated defaults
  * Expanded and enabled feature-gated tests to validate sizing and related flows


programs/registry/src/protocol_config/state.rs
```diff
@@ -57,7 +57,7 @@ impl Default for ProtocolConfig {
             active_phase_length: 1000,
             report_work_phase_length: 100,
             network_fee: 5000,
-            cpi_context_size: 20 * 1024 + 8,
+            cpi_context_size: 14020,
```
SwenSchaeferjohann: cant we use DEFAULT_CPI_CONTEXT_ACCOUNT_SIZE_V2 here

SwenSchaeferjohann (approved):
just one comment

PR #2035, opened 2025-11-05
fix: add compress and close account uniqueness

ananas-block:
**Issue:**
- potential attack to compress and close 2 compressible ctoken accounts to one compressed token account and steal half of the funds by declaring the same output compressed token account as output for both ctoken accounts.
- it is unlikely that the forester was able to exploit this issue `compress_and_close_ctoken_accounts_with_indices` builds the instruction based on the ctoken accounts in the registry program -> its not possible for foresters to specify compressed output accounts.
-  As defensive measure we still add a uniqueness check.

**Changes:**
- add check for output uniqueness `ErrorCode::CompressAndCloseDuplicateOutput` (incl unit test)
- refactor: move `light-registry` compress and close instruction building from compressed-token-sdk to `light-registry` (the program should not depend on sdk-libs crates)
- refactor: require one `authority_index` and `destination_index` on instruction level instead of account level 

## Summary by CodeRabbit

* **New Features**
  * Added duplicate output detection for CompressAndClose operations to prevent unintended fund routing.

* **Bug Fixes**
  * Implemented validation to reject CompressAndClose transactions referencing duplicate compressed output accounts.

* **Chores**
  * Added bitvec dependency to support enhanced validation logic.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -229,8 +229,7 @@ num-traits = "0.2.19"
 zerocopy = { version = "0.8.25" }
 base64 = "0.13"
 zeroize = "=1.3.0"
-bitvec = "1.0.1"
-
+bitvec = { version = "1.0.1", default-features = false }
 # HTTP client
 reqwest = "0.12"
 
--- a/program-libs/bloom-filter/Cargo.toml
+++ b/program-libs/bloom-filter/Cargo.toml
@@ -11,7 +11,7 @@ solana = ["dep:solana-program-error"]
 pinocchio = ["dep:pinocchio"]
 
 [dependencies]
-bitvec = "1.0.1"
+bitvec  = { workspace = true }
 solana-nostd-keccak = "0.1.3"
 num-bigint = { workspace = true }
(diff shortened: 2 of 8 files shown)


programs/compressed-token/program/src/transfer2/compression/ctoken/compress_and_close.rs
```diff
@@ -168,33 +168,58 @@ fn validate_compressed_token_account(
 /// Close ctoken accounts after compress and close operations
 pub fn close_for_compress_and_close(
     compressions: &[ZCompression<'_>],
-    validated_accounts: &Transfer2Accounts,
+    _validated_accounts: &Transfer2Accounts,
```
SwenSchaeferjohann, line 171: can the param be removed now?

ananas-block: it can't it's undescored to suppress the unused warning during `cargo test`, we still use  `_validated_accounts` in target os solana (the called function uses svm syscalls that don't work in my unit test).

PR #2045, opened 2025-11-10
fix: batched merkle tree zeroing

ananas-block:
**Issue:**
- current implementation zeroes all unsafe roots except for the last one
- `batch.root_index` marks the last unsafe root not the first safe root

**Fix:**
- zero out all unsafe roots `1..num_remaining_roots {` -> `0..num_remaining_roots {`
- add additional constraint that one safe root must exist after the last unsafe root so that we don't zero out all roots.
- add check that we do not zero out the complete root history (and delete the tree).

**Changes:**
- create `fn should_zero_bloom_filter`, refactors conditional logic into a separate function for easier formal verification 
- refactor `batched-merkle-tree-tests/merkle_tree.rs` into individual files in a directory
- extend asserts in merkle tree tests to check that no unsafe root remains post zeroing operation

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unsafe clearing of root history by introducing an explicit error that is now returned and propagated during root-zeroing operations.

* **Tests**
  * Expanded and reorganized end-to-end tests for address, state, and transaction flows; added focused modules exercising batching, bloom filters, and root-history scenarios.
  * Removed an old monolithic test suite and tweaked several test iteration counts for more accurate scenarios.

* **Chores**
  * Added a dev test dependency to support new test modules.

### DIFF
--- a/program-libs/batched-merkle-tree/src/merkle_tree.rs
+++ b/program-libs/batched-merkle-tree/src/merkle_tree.rs
@@ -1278,10 +1278,6 @@ mod test {
 
             assert_ne!(previous_roots, current_roots);
             let root_index = account.queue_batches.batches[0].root_index;
-            // assert_eq!(
-            //     account.root_history[root_index as usize],
-            //     previous_roots[root_index as usize]
-            // );
 
             assert_eq!(
                 account.queue_batches.batches[0].get_state(),
@@ -1538,7 +1534,6 @@ mod test {
                     .queue_batches
                     .increment_pending_batch_index_if_inserted(state);
             }
-            //last_batch1_root_update2 = account.root_history.last().unwrap();
             assert_eq!(
                 account.queue_batches.batches[0].get_state(),
                 BatchState::Inserted


program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -1214,23 +1236,53 @@ mod test {
...
             assert_ne!(previous_roots, current_roots);
             let root_index = account.queue_batches.batches[0].root_index;
-            assert_eq!(
-                account.root_history[root_index as usize],
-                previous_roots[root_index as usize]
-            );
+            // assert_eq!(
+            //     account.root_history[root_index as usize],
+            //     previous_roots[root_index as usize]
+            // );
```
SwenSchaeferjohann: does the assert need to be enabled?

ananas-block: no

ananas-block:
```suggestion
```

program-libs/batched-merkle-tree/src/merkle_tree.rs
```diff
@@ -1451,6 +1538,7 @@ mod test {
                     .queue_batches
                     .increment_pending_batch_index_if_inserted(state);
             }
+            //last_batch1_root_update2 = account.root_history.last().unwrap();
```
ananas-block:
```suggestion
```

SwenSchaeferjohann (approved):

PR #2049, opened 2025-11-11
refactor: remove bump from create ata instruction data

ananas-block:
**Issue:**
- create ata passes a bump which allows to create multiple cmint accounts from a single seed provided multiple bumps that result in an offcurve pda exist.

**Changes:**
- remove mint_bump from create mint instruction data

## Release Notes

* **Refactor**
  * Simplified Associated Token Account creation API by removing bump parameter variants and consolidating functions into a unified approach.
  * Moved PDA verification logic to occur internally during account creation rather than through separate validation steps.

* **Tests**
  * Updated tests to validate deterministic ATA address derivation across different owners and mints.


SwenSchaeferjohann (approved):
LGTM

ananas-block:
too CU intensive

PR #2052, opened 2025-11-12
feat: forester compressible ctoken support

ananas-block:
**Changes:**
- regenerate accounts and add new test-validator env accounts:
    1. rent sponsor
    2. compressible config pda
    3. config counter pda
- add compressible flag to forester start
- add `Compressor` compresses and closes any compressible ctoken accounts that have been indexed and saved into `CompressibleAccountTracker`
- add `AccountSubscriber` subscribes to account changes of the `CToken` program and indexes compressible ctoken accounts
- add `get_multiple_accounts` to `trait Rpc`

**Note:**
- in a different pr we should replace `CompressibleAccountTracker` with a DB connection

## Summary by CodeRabbit

* **New Features**
  * Compressible token support: discovery (bootstrap), account tracking, and automated batch compression with concurrency controls.
  * Real-time WebSocket subscriber to keep the tracker up to date.
  * Batch RPC account retrieval to fetch multiple accounts in one call.

* **Configuration**
  * New --enable-compressible toggle (off by default) plus compressible WS URL, batch size, and max concurrent batches.
  * Separate shutdown channels for graceful compressible/bootstrap teardown.

* **Tests**
  * New and extended end-to-end tests covering bootstrap, tracking, and compression.

* **Bug Fixes**
  * Safe handling for unknown/virtual tree types to avoid runtime errors.

* **Chores**
  * Added workspace dependencies to support the compressible feature.

### DIFF
--- a/forester/src/compressible/bootstrap.rs
+++ b/forester/src/compressible/bootstrap.rs
@@ -17,7 +17,7 @@ use crate::Result;
 const PAGE_SIZE: usize = 10_000;
 
 /// Bootstrap the compressible account tracker by fetching existing accounts
-/// Uses standard getProgramAccounts for localhost, getProgramAccountsV2 for mainnet
+/// Uses standard getProgramAccounts for localhost, getProgramAccountsV2 for remote networks
 pub async fn bootstrap_compressible_accounts(
     rpc_url: String,
     tracker: Arc<CompressibleAccountTracker>,
@@ -31,7 +31,7 @@ pub async fn bootstrap_compressible_accounts(
         info!("Detected localhost, using standard getProgramAccounts");
         bootstrap_with_standard_api(rpc_url, tracker, shutdown_rx).await
     } else {
-        info!("Using Helius getProgramAccountsV2 with pagination");
+        info!("Using getProgramAccountsV2 with pagination");
         bootstrap_with_v2_api(rpc_url, tracker, shutdown_rx).await
     }
 }
@@ -68,10 +68,18 @@ fn process_account(
         }
     };
 
-    let lamports = account_obj
-        .get("lamports")
-        .and_then(|v| v.as_u64())
-        .unwrap_or(0);
+    // Check lamports - skip closed accounts (lamports == 0)
+    let lamports = match account_obj.get("lamports").and_then(|v| v.as_u64()) {
+        Some(0) => {
+            debug!("Skipping closed account {} (lamports == 0)", pubkey);
+            return Ok(false);
+        }
+        Some(lamports) => lamports,
+        None => {
+            debug!("Skipping account {} with missing lamports field", pubkey);
+            return Ok(false);
+        }
+    };
 
     let data_array = match account_obj.get("data").and_then(|v| v.as_array()) {
         Some(arr) if !arr.is_empty() => arr,
@@ -208,7 +216,7 @@ async fn bootstrap_with_v2_api(
 
         // Add cursor for pagination
         if let Some(ref c) = cursor {
-            params[1]["cursor"] = json!(c);
+            params[1]["paginationKey"] = json!(c);
         }
 
         let payload = json!({
@@ -274,7 +282,7 @@ async fn bootstrap_with_v2_api(
 
         // Get cursor for next page
         cursor = result
-            .get("cursor")
+            .get("paginationKey")
             .and_then(|c| c.as_str())
             .map(|s| s.to_string());
 
--- a/forester/src/compressible/state.rs
+++ b/forester/src/compressible/state.rs
@@ -7,38 +7,45 @@ use light_ctoken_types::{
     COMPRESSIBLE_TOKEN_ACCOUNT_SIZE, COMPRESSIBLE_TOKEN_RENT_EXEMPTION,
 };
 use solana_sdk::pubkey::Pubkey;
-use tracing::debug;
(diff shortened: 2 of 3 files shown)


cli/accounts/batched_output_queue_oq1na8gojfdUhsfCpyjNt6h4JaDWtHf1yQj4koBWfto.json
SwenSchaeferjohann, line 1: I cant see the difference?

forester/src/compressible/state.rs
```diff
@@ -0,0 +1,166 @@
...
+                    })
+                {
+                    // Get the last funded epoch using the extension's method
+                    let last_funded_epoch = compressible_ext
+                        .get_last_funded_epoch(
+                            COMPRESSIBLE_TOKEN_ACCOUNT_SIZE,
+                            state.lamports,
+                            COMPRESSIBLE_TOKEN_RENT_EXEMPTION,
+                        )
+                        .unwrap_or(0);
```
SwenSchaeferjohann: if the unwrap returns an error this will return epoch = 0 and try to compress, no? should we handle this explicitly

forester/src/compressible/state.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// Update account state from raw account data
+    pub fn update_from_account(
+        &self,
+        pubkey: Pubkey,
+        account_data: &[u8],
+        lamports: u64,
+    ) -> Result<()> {
+        // Deserialize CToken using borsh
+        let ctoken = CToken::try_from_slice(account_data)
+            .map_err(|e| anyhow::anyhow!("Failed to deserialize CToken with borsh: {:?}", e))?;
```
SwenSchaeferjohann, line 123: can we be certain that account will only ever be of type CToken?

forester/src/compressible/state.rs
```diff
@@ -0,0 +1,166 @@
...
+    ) -> Result<()> {
+        // Query all accounts at once using get_multiple_accounts
+        let accounts = rpc.get_multiple_accounts(pubkeys).await?;
+
+        for (pubkey, account_opt) in pubkeys.iter().zip(accounts.iter()) {
+            match account_opt {
+                Some(account) => {
+                    // Account exists - update lamports
+                    if let Some(mut state) = self.accounts.get_mut(pubkey) {
+                        state.lamports = account.lamports;
```
SwenSchaeferjohann: is there a reason why we only update lamports but not the account state here? would that not result in stale token data if it had changes since last sync?

forester/src/compressible/subscriber.rs
```diff
@@ -0,0 +1,147 @@
...
+            response.value.account.lamports,
+        ) {
+            Ok(()) => {
+                debug!(
+                    "Updated account {} at slot {}",
+                    pubkey, response.context.slot
+                );
+            }
+            Err(e) => {
+                debug!("Failed to update tracker for {}: {}", pubkey, e);
```
sergeytimoshin:
```suggestion
                error!("Failed to update tracker for {}: {}", pubkey, e);
```
+ probably makes sense to add err propagation otherwise we could just ignore it silently

forester/src/epoch_manager.rs
```diff
@@ -1616,6 +1641,90 @@ impl<R: Rpc> EpochManager<R> {
...
+            debug!(
+                "Processing compression batch {}/{} with {} accounts",
+                batch_idx + 1,
+                num_batches,
+                batch.len()
+            );
+
+            match compressor
+                .compress_batch(batch, registered_forester_pda)
+                .await
```
sergeytimoshin: how it works on the program side? can we parallelize batch compression?

ananas-block: yes we can

forester/src/compressible/state.rs
ananas-block: The user has is no way to update the compressible extension data.
The claim could update the rent config but that is currently unimplemented as well.

forester/src/compressible/state.rs
ananas-block: We currently filter by account size.

cli/accounts/batched_output_queue_oq1na8gojfdUhsfCpyjNt6h4JaDWtHf1yQj4koBWfto.json
ananas-block: I think the length of the cpi context account changed -> additional bytes field in tree and queue metadata.

forester/src/compressible/bootstrap.rs
```diff
@@ -0,0 +1,367 @@
...
+            info!(
+                "Bootstrap complete: reached end of results at page {}",
+                page_count
+            );
+            break;
+        }
+
+        // Add small delay between requests to avoid rate limiting
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+    }
```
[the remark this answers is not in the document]
ananas-block: ✅ **Resolved** - Changed pagination field from `cursor` to `paginationKey` and added graceful fallback to standard API when V2 is not supported.

forester/src/epoch_manager.rs
ananas-block: done

PR #2053, opened 2025-11-13
fix: top up or rent transfer mint to ctoken

ananas-block:
## Summary by CodeRabbit

* **Bug Fixes**
  * Stronger overflow checks on mint amounts with clearer errors to prevent transfer failures during compressed-token mints.

* **Refactor**
  * Per-account transfer accumulation and batched lamport transfers with a capped number of transfer entries for deterministic processing and fewer individual transfers.

* **Breaking Changes**
  * Mint action payload flattened to explicit account index and amount fields; mint operations may now return transfer results to callers (update integrations accordingly).

* **Tests**
  * Updated test vectors and generators to match the simplified mint action payload.

### DIFF
--- a/programs/compressed-token/program/src/mint_action/actions/process_actions.rs
+++ b/programs/compressed-token/program/src/mint_action/actions/process_actions.rs
@@ -30,6 +30,9 @@ use crate::{
     },
 };
 
+/// Maximum number of packed accounts allowed in a single instruction
+const MAX_PACKED_ACCOUNTS: usize = 40;
+
 #[allow(clippy::too_many_arguments)]
 #[profile]
 pub fn process_actions<'a>(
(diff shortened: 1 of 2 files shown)


programs/compressed-token/program/src/mint_action/actions/process_actions.rs
```diff
@@ -80,13 +91,27 @@ pub fn process_actions<'a>(
...
                     mint_to_ctoken_action,
                     compressed_mint,
                     validated_accounts,
                     packed_accounts,
                     parsed_instruction_data.mint.metadata.mint,
                 )?;
+
+                // Accumulate transfer amount if present (deduplication happens here)
+                if let Some((account_index, amount)) = transfer {
+                    if account_index as usize > 40 {
```
SwenSchaeferjohann: off by one
```suggestion
                    if account_index as usize >= 40 {
```

sergeytimoshin (approved):

PR #2064, opened 2025-11-20
refactor: compressed token sdk transfer interface builder

ananas-block:
**Changes:**
- implement the following pattern for `CTokenTransfer` and `TransferInterface`

**Pattern:**
```
pub struct OperationName {
      pub account1: Pubkey,
      pub account2: Pubkey,
      pub amount: u64,
      // ... other fields
  }

  impl OperationName {
      pub fn instruction(self) -> Result<Instruction, ProgramError> {
          // Build and return Instruction
      }
  }

pub struct OperationNameAccountInfos<'info> {
      pub account1: AccountInfo<'info>,
      pub account2: AccountInfo<'info>,
      pub amount: u64,
      // ... other fields (matching Data struct)
  }

  impl<'info> OperationNameAccountInfos<'info> {
      pub fn instruction(&self) -> Result<Instruction, ProgramError> {
          OperationName::from(self).instruction()
      }

      pub fn invoke(self) -> Result<(), ProgramError> {
          let instruction = OperationName::from(&self).instruction()?;
          let account_infos = [self.account1, self.account2, ...];
          invoke(&instruction, &account_infos)
      }

      pub fn invoke_signed(self, signer_seeds: &[&[&[u8]]]) -> Result<(), ProgramError> {
          let instruction = OperationName::from(&self).instruction()?;
          let account_infos = [self.account1, self.account2, ...];
          invoke_signed(&instruction, &account_infos, signer_seeds)
      }
  }
```

## Summary by CodeRabbit

* **Refactor**
  * Updated compressed token transfer API from function-based to struct-based builders
  * Reorganized transfer abstractions for ctoken and SPL token conversion operations
  * Introduced unified transfer interface supporting multiple token conversion paths with optional configuration
  * Centralized instruction construction and invocation patterns across transfer operations

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/sdk-libs/compressed-token-sdk/src/instructions/transfer_interface.rs
+++ b/sdk-libs/compressed-token-sdk/src/instructions/transfer_interface.rs
@@ -7,7 +7,7 @@ use solana_instruction::{AccountMeta, Instruction};
 use solana_program_error::ProgramError;
 use solana_pubkey::Pubkey;
 
-use super::transfer_ctoken::{transfer_ctoken, transfer_ctoken_signed};
+use super::transfer_ctoken::TransferCtokenAccountInfos;
 use crate::{
     account2::CTokenAccount2,
     error::TokenSdkError,
@@ -51,7 +51,7 @@ impl<'info> TransferSplToCtokenAccountInfos<'info> {
 
     pub fn invoke(self) -> Result<(), ProgramError> {
         let instruction = TransferSplToCtoken::from(&self).instruction()?;
-        let account_infos = vec![
+        let account_infos = [
             self.payer,
             self.compressed_token_program_authority,
             self.mint,                       // Index 0: Mint
@@ -66,7 +66,7 @@ impl<'info> TransferSplToCtokenAccountInfos<'info> {
 
     pub fn invoke_signed(self, signer_seeds: &[&[&[u8]]]) -> Result<(), ProgramError> {
         let instruction = TransferSplToCtoken::from(&self).instruction()?;
-        let account_infos = vec![
+        let account_infos = [
             self.payer,
             self.compressed_token_program_authority,
             self.mint,                       // Index 0: Mint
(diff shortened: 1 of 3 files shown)


sdk-libs/compressed-token-sdk/src/instructions/transfer_interface.rs
```diff
@@ -18,598 +18,640 @@ use crate::{
...
+    /// * `compressed_token_pool_pda` - Optional token pool PDA (required for SPL<->ctoken transfers)
+    /// * `compressed_token_pool_pda_bump` - Optional bump seed for token pool PDA
+    pub fn new(
+        source_account: AccountInfo<'info>,
+        destination_account: AccountInfo<'info>,
+        authority: AccountInfo<'info>,
+        amount: u64,
+        payer: AccountInfo<'info>,
+        compressed_token_program_authority: AccountInfo<'info>,
+    ) -> Self {
```
SwenSchaeferjohann, line 480: @ananas-block 
can we have one that lets me pass these as options:
	•	mint: Option<>
	•	c_token_pool_pda: Option<>
	•	c_token_pool_pda_bump: Option<>

this makes it easiest for AMMs because then they just need to forward the options from the instruction accounts struct.

the Default/most common usage of `TransferInterface` is that the caller programs don't know at compile time which individual bridge param will be `Some` or `None`.

we can still keep the params explicit, eg with something like this:

```rust
.new(..., mint_option) // i think adding mint_option to new makes sense
.token_pool(c_token_pool_pda_option, c_token_pool_pda_bump_option) // bump none = 0
.source_token_program(source_token_program)
.dest_token_program(destination_token_program) 
.invoke(...)
```
or 
```rust
.new(..., mint_option) // i think adding mint_option to new makes sense
.token_pool(c_token_pool_pda_option, c_token_pool_pda_bump_option) // bump none = 0
.token_programs(source_token_program, dest_token_program)
.invoke(...)
```

(note that later we will enable SPL<>SPL so we must always pass source and dest token programs)

ananas-block: at first glance sounds good to me.
I am adding tests in https://github.com/Lightprotocol/light-protocol/pull/2066 it makes more sense to merge this one and add it there so that I can test it right away.

SwenSchaeferjohann (approved):

PR #2070, opened 2025-11-23
feat: state update simplified

sergeytimoshin:


forester/src/processor/v2/common.rs
```diff
@@ -363,134 +337,43 @@ impl<R: Rpc> BatchProcessor<R> {
...
-            self.context.merkle_tree
-        );
+
+        state::perform_state_update(&self.context, merkle_tree_data, output_queue_data).await?;
 
         let mut cache = self.context.ops_cache.lock().await;
         cache.cleanup_by_key(&batch_hash);
 
-        Ok(zkp_batch_size)
+        Ok(1)
```
ananas-block, line 374: what's this magic number?

forester-utils/src/instructions/state.rs
```diff
@@ -0,0 +1,478 @@
...
+                proofs_buffer.clear();
+            }
+        }
+
+        if !proofs_buffer.is_empty() {
+            yield Ok(BatchInstruction::Nullify(proofs_buffer));
+        }
+    };
+
+    Ok((Box::pin(stream), zkp_batch_size))
```
ananas-block, line 477: `zkp_batch_size` seems to be unused, can we remove it?

forester-utils/src/instructions/state.rs
```diff
@@ -0,0 +1,478 @@
...
+            let tx_hashes = batch_data.tx_hashes.clone();
+
+            let path_indices: Vec<u32> = batch_data.leaf_indices.iter().map(|&idx| idx as u32).collect();
+
+            let mut nullifiers = Vec::with_capacity(leaves.len());
+            for i in 0..leaves.len() {
+                let mut leaf_index_bytes = [0u8; 32];
+                leaf_index_bytes[24..].copy_from_slice(batch_data.leaf_indices[i].to_be_bytes().as_slice());
+
+                let nullifier = match Poseidon::hashv(&[&leaves[i], &leaf_index_bytes, &tx_hashes[i]]) {
```
ananas-block, line 392: Photon also does this computation when parsing the v2 event, maybe we can store and return the nullifiers from photon.

forester-utils/src/instructions/state.rs
```diff
@@ -0,0 +1,478 @@
...
+                        staging_tree = Some(tree);
+                    },
+                    Err(e) => {
+                        yield Err(ForesterUtilsError::Prover(format!("Failed to initialize staging tree: {}", e)));
+                        return;
+                    }
+                }
+            }
+
+            let staging = staging_tree.as_mut().unwrap();
```
ananas-block, line 380: can the unwrap fail?

forester-utils/src/instructions/state.rs
```diff
@@ -0,0 +1,478 @@
...
+
+#[allow(clippy::too_many_arguments)]
+pub async fn get_state_update_instruction_stream<'a, R: Rpc>(
+    rpc_pool: Arc<SolanaRpcPool<R>>,
+    merkle_tree_pubkey: Pubkey,
+    prover_append_url: String,
+    prover_update_url: String,
+    prover_api_key: Option<String>,
+    polling_interval: Duration,
+    max_wait_time: Duration,
```
ananas-block, line 44: can we bundle these into a config struct?

forester-utils/src/instructions/state.rs
```diff
@@ -0,0 +1,478 @@
...
+    (
+        Pin<Box<dyn Stream<Item = Result<BatchInstruction, ForesterUtilsError>> + Send + 'a>>,
+        u16,
+    ),
+    ForesterUtilsError,
+> {
+    let (merkle_tree_next_index, current_root, _) = (
+        merkle_tree_data.next_index,
+        merkle_tree_data.current_root,
+        merkle_tree_data.root_history,
```
ananas-block, line 58: seems to be unused

prover/client/src/proof_types/batch_update/proof_inputs.rs
```diff
@@ -134,3 +134,58 @@ pub fn get_batch_update_inputs<const HEIGHT: usize>(
 
     Ok((inputs, changelog))
 }
+
+#[allow(clippy::too_many_arguments)]
+pub fn get_batch_update_inputs_v2<const HEIGHT: usize>(
```
ananas-block, line 139: it would be more efficient to move this logic into `process_batch_updates`.

`process_batch_updates` collects all proofs and other inputs into a vector only to pass it into this function which performs only type conversions and collects all inputs into new vectors again.
We can save one vector allocation if we do the type conversion right away in `process_batch_updates`.

forester-utils/src/instructions/state.rs
```diff
@@ -0,0 +1,478 @@
...
+pub async fn get_state_update_instruction_stream<'a, R: Rpc>(
+    rpc_pool: Arc<SolanaRpcPool<R>>,
+    merkle_tree_pubkey: Pubkey,
+    prover_append_url: String,
+    prover_update_url: String,
+    prover_api_key: Option<String>,
+    polling_interval: Duration,
+    max_wait_time: Duration,
+    merkle_tree_data: ParsedMerkleTreeData,
+    output_queue_data: Option<ParsedQueueData>,
```
ananas-block, line 46: looks like we are still depending on `get_account` rpc calls for more than just debug asserts.
`append_hash_chains` from parsed queue data is used to determine how many proofs are created.

ideal flow:
1. `get_queue_elements_v2` -> return all queue elements for an input and output queue pair (or two parallel calls)
2. create proof inputs for output and input queues

I suppose the current design prioritizes the output queue over the input queue which should be fine since it is possible to create more outputs than inputs.

forester-utils/src/instructions/state.rs
```diff
@@ -0,0 +1,478 @@
...
+
+        let mut proofs_buffer = Vec::new();
+
+        for (batch_idx, leaves_hash_chain) in append_hash_chains.iter().enumerate() {
+            if !proofs_buffer.is_empty() && batch_idx > 0 {
+                yield Ok(BatchInstruction::Append(proofs_buffer.clone()));
+                proofs_buffer.clear();
+            }
+
+            let queue_elements_result = {
```
ananas-block, line 121: can you extract append and update proof gen into a function each?

PR #2091, opened 2025-12-03
feat: add `transfer2WithAta` ixn wrapper

SwenSchaeferjohann:
## Description
* adds a wrapper ixn `transfer2WithAta` that derives the ATA and invokes transfer2 with the ata a signer via self-cpi

## why
improving devex for the same reasons why solana had added atas back in the day.

## alternative considered

1) **not adding.** devex improvement too big to ignore for payments companies and trading apps.

2) **not doing self-cpi but call process_transfer2 and change transfer2 ownership check**. saves CU cost but adds audit complexity. whereas the self-cpi approach is purely accretive. 

self-cpi adds ~4400 cu extra cost for decompressing 1 ATA compared to directly invoking transfer2 with proveByIndex:true.

I think the extra cpi cost is still worth it also because:
- creating ctoken atas with compressToPubkey=true is optional, so power users can still choose to create it the non-ata way
- this is outside the on the cold path, so the cu cost is less critical
- the devex improvements outweigh these drawbacks

## test coverage

* Success
test_transfer2_with_ata_single_input_success - Decompress single ATA-owned compressed token
test_transfer2_with_ata_multiple_inputs_success - Decompress multiple ATA-owned tokens in one call
* Failing
test_transfer2_with_ata_wrong_owner_signer_fails - Wrong wallet (not ATA owner) tries to sign
test_transfer2_with_ata_wrong_mint_fails - Mint doesn't match ATA derivation
test_transfer2_with_ata_wallet_not_signer_fails - Correct wallet key but missing signature
test_transfer2_with_ata_wrong_ata_account_fails - Wrong ATA address passed

test_transfer2_with_ata_mixed_ownership_fails - Mix of ATA-owned and wallet-owned inputs (must all be ATA-owned)
note this is implementation detail to reduce complexity, we could remove this check safely.

test_transfer2_with_ata_wrong_bump_fails - bump must match
test_transfer2_with_ata_cu_benchmark - CU comparison vs regular Transfer2 (1 & 2 inputs)

## changes to existing program code:

* at ctoken creation, do not disable `compressToPubkey` for ATAs, but reuse check_seeds() to keep diff minimal: 

```rust
    if let Some(compress_to_pubkey) = compressible_config_ix_data
        .compress_to_account_pubkey
        .is_some()
    {
        msg!("Associated token accounts must not compress to pubkey");
        return Err(ProgramError::InvalidInstructionData);
    }
```
becomes: 
```rust
    if let Some(compress_to_pubkey) = compressible_config_ix_data
        .compress_to_account_pubkey
        .as_ref()
     compress_to_pubkey
            .check_seeds(associated_token_account.key())
            .map_err(|_| {
                msg!("compress_to_pubkey seeds do not match ATA derivation");
                ProgramError::InvalidInstructionData
            })?;
```

## cu overhead comparison

1 input(s) with ata    : 44363 cu
1 input(s) transfer2   : 39142 cu
1 input(s) difference  : +5221 cu (+13.3%)

2 input(s) with ata    : 56460 cu
2 input(s) transfer2   : 46849 cu
2 input(s) difference  : +9611 cu (+20.5%)

## additional changes we could consider:
* could safely remove the check that enforces all compressed token accounts to be owned by ATA. (kept for simple audit overhead)
* could tighten this ixn to just allow decompression flow and check that the destination key again matches the ATA (this would simplify usage but restrict transfer2WithAta unnecessarily)

## Release Notes

* **New Features**
  * Added support for transfers of compressed tokens owned by associated token accounts (ATAs).
  * Introduced a new Transfer2WithAta instruction type for ATA-based token operations.
  * Added utilities for decompressing ATA-owned compressed tokens with optional delegate support.

* **Improvements**
  * Enhanced configuration propagation for associated token account creation.
  * Expanded multi-signer support in transaction simulation tooling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/program-tests/compressed-token-test/tests/transfer2/transfer2_with_ata.rs
+++ b/program-tests/compressed-token-test/tests/transfer2/transfer2_with_ata.rs
@@ -332,7 +332,7 @@ async fn test_transfer2_with_ata_single_input_success() {
 
     // Verify ATA now has the tokens
     let ata_account = ctx.rpc.get_account(ctx.ata).await.unwrap().unwrap();
-    assert!(ata_account.data.len() > 0, "ATA should exist with tokens");
+    assert!(!ata_account.data.is_empty(), "ATA should exist with tokens");
 }
 
 /// Test: Successfully decompress multiple ATA-owned compressed tokens in single call
@@ -380,7 +380,7 @@ async fn test_transfer2_with_ata_multiple_inputs_success() {
 
     // Verify all tokens are now in ATA
     let ata_account = ctx.rpc.get_account(ctx.ata).await.unwrap().unwrap();
-    assert!(ata_account.data.len() > 0, "ATA should exist with tokens");
+    assert!(!ata_account.data.is_empty(), "ATA should exist with tokens");
 }
 
 // ============================================================================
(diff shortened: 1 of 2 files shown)


ananas-block (commented):
Interesting solution, let's hold off on merging this until we have all audit fixes for the current state merged so that we can split a clean release branch.

PR #2363, opened 2026-03-27
refactor: bump anchor & solana crates

sergeytimoshin:


external/anchor
ananas-block, line 1: why?

program-libs/account-checks/src/account_info/test_account_info.rs
```diff
@@ -5,66 +5,62 @@ pub mod pinocchio {
     extern crate std;
     use std::vec::Vec;
 
-    use pinocchio::{account_info::AccountInfo, instruction::Account, pubkey::Pubkey};
+    use pinocchio::{account::RuntimeAccount, AccountView as AccountInfo};
     use rand::{prelude::Rng, thread_rng};
 
-    pub fn pubkey_unique() -> Pubkey {
+    pub fn pubkey_unique() -> [u8; 32] {
```
ananas-block: maybe it is redundant now if we use address

PR #2384, opened 2026-06-19
fix: import sys_hash for SBPF v3 static syscalls in light-hasher

saigowthamch22:
## Summary

- `light-hasher` 5.0.0 copies Solana's `define_syscall!` macro for static syscalls but never defines or imports `sys_hash`, causing compile failures when building with SBPF v3 (`--arch v3`, required for new deploys under SIMD-0500).
- Adds `solana-define-syscall` and re-exports `sys_hash` under `#[cfg(target_feature = "static-syscalls")]`, matching `solana-program` 2.3.

## Reproduction

Before this fix, building with SBPF v3 fails:

```
error[E0425]: cannot find function `sys_hash` in this scope
  --> program-libs/hasher/src/syscalls/definitions.rs:13:12
```

## Test plan

- [x] `cargo build-sbf --arch v3` in `program-libs/hasher` succeeds locally
- [ ] CI passes
- [ ] Request patch release of `light-hasher` on crates.io (e.g. 5.0.1) so downstream crates like `light-sdk` consumers can build with SBPF v3 without `[patch.crates-io]`

## Chores
* Added new workspace dependency to improve syscall handling capabilities.
* Updated syscall definitions to support static-syscalls build configuration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4011,6 +4011,7 @@ dependencies = [
  "rand 0.8.5",
  "sha2 0.10.9",
  "sha3",
+ "solana-define-syscall 2.3.0",
  "solana-program-error",
  "solana-pubkey",
  "thiserror 2.0.18",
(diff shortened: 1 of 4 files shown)


ktsourdi-sbt (approved):
faced the same issue, thats the correct solution

PR #2389, opened 2026-08-24
fix: write only defined bytes into concurrent Merkle tree changelog

ananas-block:
## Problem

`ConcurrentMerkleTree` writes changelog entries into the account buffer via `CyclicBoundedVec::push`, which does a `ptr::write` of a `ChangelogEntry` built on the stack. Two regions of that struct have no defined value on the stack and were copied verbatim into the account:

- the 32 value bytes of every `None` node in `ChangelogPath` (`Option<[u8; 32]>` only defines the tag byte for `None`), and
- the `repr(C)` padding between `path` (`33 * HEIGHT` bytes, align 1) and `index: u64` (2 bytes at height 22, 6 bytes at height 26).

Account bytes therefore depended on runtime stack contents, which is the source of the ok/ok divergence observed when toggling the virtual address space feature (see the `mollusk-vas-repro`).

## Fix

All three changelog pushes (`init`, `update_leaf_in_tree`, `append_batch`) go through a new private `push_changelog_entry`, which zeroes the slot and then writes only `index` and the `Some` nodes. `None` nodes and the padding stay zero.

No type, layout, or public API change. Existing accounts remain readable; the only observable difference is that previously undefined bytes are now zero.

A `const` assertion pins the `Option<[u8; 32]>` size the helper relies on.

## Tests

- `test_changelog_layout`: node is 33 bytes with tag 0/1 at byte 0; `ChangelogEntry` sizes and `index` offsets for heights 22/26/32/40 (736/872/1064/1328).
- `test_changelog_bytes_are_defined`: `0xFF`-prefilled buffer, `init` + `append_batch` at height 10 (6 padding bytes); asserts every node is `tag == 1` or all-zero and the padding is zero.

Verified: `cargo test -p light-concurrent-merkle-tree`, clippy `-D warnings`, `cargo check` of `light-test-utils` / `account-compression` / `light-indexed-merkle-tree` / `forester`, and `cargo build-sbf` of `account-compression`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of changelog records during tree initialization, updates, and batch operations.
  * Ensured consistent serialized data and memory layouts for Merkle tree and hash set storage.
  * Improved hash set initialization, insertion, and buffer-size validation, including clearer handling of reserved storage.

* **Tests**
  * Added coverage for deterministic serialization, complete byte initialization, storage sizing, field placement, and state transitions across supported data structures.


ananas-block:
Superseded by #2392, which contains only the program-libs changes, squashed and rebased on main, plus the CI fix for the dead-code error.

PR #2391, opened 2026-09-23
fix(forester): bound late proof cache warming

sergeytimoshin:
## Summary
- cap late-proof collection at 30 seconds
- always release the per-tree cache warming state after channel closure or timeout
- add a regression test for a proof sender that remains alive

## Production symptom
Devnet V2 processing eventually stopped while the processes remained healthy. Every eligible slot was deferred because the proof cache stayed in warming state indefinitely. Restarting cleared the in-memory state but only recovered processing temporarily.

## Validation
- cargo test -p forester processor::v2::tx_sender::tests
- cargo clippy -p forester --lib -- -D warnings
- cargo fmt --all -- --check

Depends on #2387.


sergeytimoshin:
Follow-up fix: 1ecdb0a2471bf97900c1af9fd02265901f082c4b.

Ready proof prefixes can now be sent while collection continues; cache entries are acknowledged only after confirmation. Cached-send errors remain typed so ForesterNotEligible triggers schedule recovery. Pending collections suppress duplicate speculative chains, out-of-order handoff retains proofs across gaps, and prover payloads no longer include the changing local sequence number. Cached-work metrics now count queue items rather than proof instructions.

Validation: 70 forester library tests passed; cargo clippy -p forester --lib -- -D warnings passed; forester formatting and release build passed.

Devnet canary, 2026-09-24 17:22:41–17:25:37 UTC: updated devnet-1 confirmed 5 cached-proof transactions / 2,500 queue items with zero cached-send failures and zero warming deferrals. Unchanged devnet-2 confirmed no processing transactions in the same interval and logged 84 warming deferrals. This is a short operational canary, not a controlled sustained-throughput benchmark.

Promoted devnet-2 after the canary; both APIs healthy with zero container restarts on the final image. Devnet-2 subsequently confirmed a 500-item batch. Mainnet and the shared prover were not restarted. An initial host-binary/runtime glibc mismatch was rolled back and corrected before the successful canary; final image: light-forester:ready-proof-cache-1ecdb0a24-noble.

